PR #2, opened 2015-05-26
Cleanups, performance, and code clarify improvements.

joshgoebel:
Lots of little performance, memory and code clarity optimizations.  See individual commits for details.


joshgoebel:
If it's better I can rip off the last few commits and make a sep pull request once the first bits are merged.  Just let me know.  Right now it's just easier to build on the same foundation.

PR #5, opened 2015-05-27
add drawSlowXYBitmap()

joshgoebel:


joshgoebel:
This is build on my prior work (since a lot has changed) but if the cleanups branchs gets merged I'll rebase this so it only includes the last two commits (the drawSlowXYBitmap stuff).

PR #6, opened 2015-05-27
fix ArduBreakout example to compile until we fix tone

joshgoebel:


joshgoebel:
Sorry I didn't realize my audio stuff was in the with graphics feature.  Poor branch management, sorry.  That wasn't 100% ready for merge - hence the tone stuff still being an issue.

This fixes it so the sketch again compiles by calling the new audio.tone() method which doesn't yet work.

PR #7, opened 2015-05-28
give contributors credit and add LICENSE

joshgoebel:


joshgoebel:
I can slap Chris's actually name on there if someone will tell me his last name.

PR #12, opened 2015-06-02
misc improvements and fixes

joshgoebel:


joshgoebel:
I can rip out TODO if I must, or we can discuss and make it official.  I found it useful to keep track of it as I work don this.  If there are any questions, please ask.  It would be nice if at least one other person compiled samples before we merged items.

joshgoebel:
This is what I've been working on while we didn't have an official repository.  After this chunk I can go back to individual pull requests for sets of feature and try to group things better.

joshgoebel:
What needs to be done here?

PR #17, opened 2015-06-06
add missing character in font table

joshgoebel:
Reference:
https://github.com/adafruit/Adafruit-GFX-Library/issues/22


joshgoebel:
Any issue with this that I need to fix to get merged?

PR #21, opened 2015-06-11
getBuffer() can't be inline if we actually want it usable

joshgoebel:


joshgoebel:
This never should have been inline.  Seems to prevent it from being called externally which defeats the entire purpose.

joshgoebel:
As the getBuffer issue is a bug, can it please be merged?

joshgoebel:
Ping.  I'd be more than happy to directly help with merges and things if I had the permission to do so.  It might be helpful for there to be a (slightly?) larger team managing contributions.  It does seem bad when one liner bug fixes take over one week to be looked at.

joshgoebel:
It would also be really nice to start using github labels... I've taken to the [label] convention since I don't have permission to label things properly.

joshgoebel:
ping

joshgoebel:
Ping.

PR #23, opened 2015-06-12
Hello, World! example

dragula96:
smallest program example.

### DIFF
--- a/examples/Buttons/Buttons.ino
+++ b/examples/Buttons/Buttons.ino
@@ -1,0 +1,105 @@
+/*
+Buttons example
+June 11, 2015
+Copyright (C) 2015 David Martinez
+All rights reserved.
+
+This code is the most basic barebone code for showing how to use buttons in Arduboy.
+
+This library is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
+License as published by the Free Software Foundation; either
+version 2.1 of the License, or (at your option) any later version.
+*/
+
+#include <SPI.h>
+#include "Arduboy.h"
+#include <EEPROM.h>
+
+// make an instance of arduboy used for many functions
+Arduboy arduboy;
+
+// Variables for your game go here.
+char text[] ="Press Buttons!";
+byte x;
+byte y;
+
+
+// This function runs once in your game.
+// use it for anything that needs to be set only once in your game.
+void setup()
+{
+  //initiate arduboy instance
+  arduboy.start();
+
+  //here we set the framerate to 30, we do not need to run at default 60 and it saves us battery life.
+  arduboy.setFrameRate(30);
+  
+  //set x and y to the middle of the screen
+  x = 24;
+  y = 32;
+
+}
+
+// our main game loop, this runs once every cycle/frame.
+// this is where our game logic goes.
+void loop()
+{
+  // pause render until it's time for the next frame
+  if (!(arduboy.nextFrame()))
+    return;
+
+
+
+// the next couple of lines will deal with checking if the D-pad buttons are pressed
+// and move our text accordingly.
+
+// if the right button is pressed move 1 pixel to the right every frame
+if(arduboy.pressed(RIGHT_BUTTON)){
+  x = x + 1;
+  
+}
+
+// if the left button is pressed move 1 pixel to the left every frame
+if(arduboy.pressed(LEFT_BUTTON)){
+  x = x - 1;
+}
+
+// if the up button is pressed move 1 pixel up every frame
+if(arduboy.pressed(UP_BUTTON)){
+  y = y - 1;
+}
+
+// if the down button is pressed move 1 pixel down every frame
+if(arduboy.pressed(DOWN_BUTTON)){
+  y = y + 1;
+}
+// if the A button is pressed move 1 pixel up every frame
+if(arduboy.pressed(A_BUTTON)){
+   y = y - 1;
+}
+
+// if the B button is pressed move 1 pixel down every frame
+if(arduboy.pressed(B_BUTTON)){
+   y = y + 1;
+}
+
+
+ 
+
+
+
+  //we clear our screen to black
+  arduboy.clearDisplay();
+
+  //we set our cursor x pixels to the right and y down from the top
+  arduboy.setCursor(x, y);
+
+  //then we print to screen what is stored in our text variable we declared earlyer
+  arduboy.print(text);
+
+  //then we finaly we tell the arduboy to display what we just wrote to the display.
+  arduboy.display();
+}
+
+
--- a/examples/Buttons/Buttons.ino
+++ b/examples/Buttons/Buttons.ino
@@ -45,7 +45,7 @@ void loop()
   arduboy.setCursor(5, 10);
 
   //then we print to screen with quotations "Hello World"
-  arduboy.print("Hello, World!");
+  arduboy.print(F("Hello, World!"));
 
   //then we finaly we tell the arduboy to display what we just wrote to the display.
   arduboy.display();
--- a/examples/Buttons/Buttons.ino
+++ b/examples/Buttons/Buttons.ino
@@ -1,9 +1,9 @@
 /*
-Hello, World! example
+Buttons example
 June 11, 2015
 Copyright (C) 2015 David Martinez
 All rights reserved.
-This code is the most basic barebone code for writing a program for Arduboy.
+This code is the most basic barebone code for showing how to use buttons in Arduboy.
 This library is free software; you can redistribute it and/or
 modify it under the terms of the GNU Lesser General Public
 License as published by the Free Software Foundation; either
@@ -17,6 +17,11 @@ version 2.1 of the License, or (at your option) any later version.
 // make an instance of arduboy used for many functions
 Arduboy arduboy;
 
+// Variables for your game go here.
+char text[] ="Press Buttons!";
+byte x;
+byte y;
+
 
 // This function runs once in your game.
 // use it for anything that needs to be set only once in your game.
@@ -25,8 +30,12 @@ void setup()
   //initiate arduboy instance
   arduboy.start();
 
-  //here we set the framerate to 15, we do not need to run at default 60 and it saves us battery life.
-  arduboy.setFrameRate(15);
+  //here we set the framerate to 30, we do not need to run at default 60 and it saves us battery life.
+  arduboy.setFrameRate(30);
+  
+  //set x and y to the middle of the screen
+  x = 24;
+  y = 32;
 
 }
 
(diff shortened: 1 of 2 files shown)


joshgoebel:
Did printing from progmem not work for you?

dragula96:
Nope , it gave me garbage, so I changed it back.

joshgoebel:
Ok, it's actually more complicated for progmem (so probably not worth it for this simple example):

http://www.arduino.cc/en/Reference/PROGMEM

joshgoebel:
Ping.

davidperrenoud:
You can either do:

```
arduboy.print(reinterpret_cast<const __FlashStringHelper *>(text));
```

Or simply:

```
arduboy.print(F("Hello, World!"));
```

dragula96:
ok, fixed "buttons.ino" and made modifications to Helloworld.ino to reflect requested changes to save memory :/

examples/Buttons/Buttons.ino
```diff
@@ -0,0 +1,101 @@
...
+
+
+  //we clear our screen to black
+  arduboy.clearDisplay();
+
+  //we set our cursor x pixels to the right and y down from the top
+  arduboy.setCursor(x, y);
+
+  //then we print to screen what is stored in our text variable we declared earlyer
+  arduboy.print(text);
```
joshgoebel, line 97: Should use the F trick here also.

joshgoebel:
If we clean these up (according to prior comments) and then squash the commit I don't see why we can't get this merged.

shogerr:
Yeah, a simple hello world would be nice. @dragula96 want to clean it up and submit a new PR? It can be merged for v1.0.0. Thanks : D

joshgoebel:
> Calling pressed() repeatedly is inefficient.

NONONO, use pressed... that's the API.  It's not inefficient.  Polling ALL the buttons takes NO time... and if it were really a problem we could cache the poll state in the frame code.

High level functions are better than low-level functions unless you have a GOOD reason.

MLXXXp:
> Polling ALL the buttons takes NO time...

What about code size?

F.Y.I. I searched a lot of existing sketches to see if any were using _not_pressed()_ and when doing so found many that use _getInput()_. (I only found _not_pressed()_ in my _ArduboyLife_ sketch.)

And why can't _getInput()_ be part of the API? We define the button bits, so what's the problem?

joshgoebel:
It is part of the low-level API, and it's the lowest level, but high level is better, cleaner, and easier to read.  Everyone uses `getInput()` because that used to be all there was.  I added pressed() only recently.  

```
if (pressed(DOWN_BUTTON))
// is clearer and more legible than
if (getInput() & DOWN_BUTTON)
```

You know what it is because you learned how it works.  But what is that & thing - and even if that's clear - is the down button wired HIGH or LOW when pressed, and if it's high or low when pressed is getInput handling that for me or do I need to handle that myself?   You can make a lot of assumptions, but there are tons of real questions here.  **Pressed is very, very unambiguous.**  It gets into pressed vs justPress vs held, but that's a completely different type of ambiguity.

None of this is substitute for a good button library (for real button state, like are you holding it, did you quickly press it, should i pause the game or unpause it because I don't know how long you held the button, etc...) my ArduboyExtra button class does all this.

Regarding size I dunno exactly but it's going to be terribly close... but  just think about what is actually happening (within the actual scope of our comparison)...

One method is calling a function, storing the result, returning the result, then doing a comparison... 
Other is storing an input, calling a function, comparing, returning the result... My money would be on pressed() being smaller, but it might be a toss up.  You're really just shifting around where you do the comparison.

joshgoebel:
And it gets you thinking higher level... `while held() {start charging}, when released(), {fire}`.  That's the right level of abstraction because you can write both and they are both clear, concise and easy to understand what's going on... not try writing that same code with getInput()... you can but you can't do it with two words... we need to know when it's released, so we need to know track "Was pushed", etc... so now you have timers and variables, etc...

The whole point of having these nice abstractions is so someone doesn't have to think that getInput is a bit mask that they must logical AND with a constant in order to determine a button is held, etc... they can just ask hey, "is that button pressed?", hey, "was that button held and just released?".

They can certainly go lower-level if the NEED to, but we shouldn't encourage it without merit.

MLXXXp:
> **Pressed is very, very unambiguous**

Actually, it's ambiguous in the fact that it's not obvious that you can also use it to check for multiple buttons being pressed, not only a single button. It's also not obvious that `pressed(A_BUTTON | B_BUTTON)` means **both** buttons must be pressed, not just either one. (Yes, I know you can also write it as `pressed(A_BUTTON + B_BUTTON)`, which is clearer to those without a firm concept of bit operations.)

If you rename _getInput()_ to _getButtonState()_ or maybe just _buttonState()_ then its purpose becomes pretty clear as well. So as long as it's documented that it returns something that can be tested using the _#define_d button bit values, I don't see why it can't be sanctioned for higher level use, when _pressed()_ or _notPressed()_ are less suitable. _pressed()_, _notPressed()_, and _buttonState()_ all deal with the same button bit _#define_s.

Also, using _pressed()_, you can't test if **only** a certain button or buttons are currently pressed while all others are released. But you could use `if (buttonState() == UP_BUTTON)`. Otherwise, we need an _onlyPressed()_ function.

> The whole point of having these nice abstractions is so someone doesn't have to think that getInput is a bit mask that they must logical AND with a constant in order to determine a button is held, etc... they can just ask hey, "is that button pressed?", hey, "was that button held and just released?".

Agreed, I don't recommend using _getinput()/buttonState()_ with logical operators when _pressed()_ or _notPressed()_ will work.  My example didn't do that. I used it with a _switch/case_ statement, to read the button state once, and then check for individual buttons one at a time. No logical operators were used. I don't think it appears as using a bit mask any more than using `if (pressed(UP_BUTTON + DOWN_BUTTON)` does. They can just say "get what button(s) are pressed", and then ask hey, "is that button pressed?",  no? then hey "is this other button pressed?" no? then hey, "how about this one?".

joshgoebel:
>  I used it with a switch/case statement, to read the button state once, and then check for individual buttons one at a time.

You can't do that, it's a byte of state information.  Your code will fail if more than one button is pressed because the full state won't match a single button.

MLXXXp:
> Your code will fail if more than one button is pressed because the full state won't match a single button.

And why isn't that acceptable in this situation? It's obvious that the user is only expected to press one button at a time for the _pollButtons_ sketch. It's slightly different behaviour when multiple buttons are pressed, but I believe it still satisfies the intent. 

How do you know my way isn't how the author would have wanted it to work if there had been an easy way to do it without using _getInput()_. Without using _getInput()_, how would you code it if you wanted each test to check for a given button pressed **regardless** if one or more other buttons are pressed at the same time? E.g. "I don't care if the _run forward_ or the _jump_ button is being held when the _fire_ button is pressed".

joshgoebel:
> And why isn't that acceptable in this situation?

It might be but for some people it's going to be an unexpected side effect - or not, exactly my point... you're guessing at how it works because the intent isn't fully expressed.

> Without using getInput(), how would you code it if you wanted each test to check for a given button pressed regardless if one or more other buttons are pressed at the same time?

That's exactly what `pressed()` does now.  It tells you if the button is pressed, period.  It only cares about the button or button mask passed to it.

joshgoebel:
Honestly case is just the wrong code there without a filter function on button... it makes the coder assume getInput is returning a list of possible exclusive buttons, not a bit mask.  So the case is the most confusing thing about that code.

joshgoebel:
You didn't go far enough to see if you put a heavily commented section at the end of the switch to explain this mess to anyone reading it - and a fall thru case to handle multi-button input , so hard to discuss it to much more based on only the first few lines. :-)

MLXXXp:
> > Without using getInput(), how would you code it if you wanted each test to check for a given button pressed regardless if one or more other buttons are pressed at the same time?
> 
> That's exactly what pressed() does now. It tells you if the button is pressed, period. It only cares about the button or button mask passed to it.

I apologise, you're right. I got confused and started thinking that _pressed()_ checked for **only** and **exactly** the specified button(s) being pressed.
I.e. `return button_state == buttons;`
not the actual `return (button_state & buttons) == buttons;`

If I wanted to do what  (in my confused state) I erroneously thought _pressed()_ did, I could use
`getInput == UP_BUTTON;`

Which kind of proves my point. I guessed at how _pressed()_ works because the intent wasn't fully expressed. The operation of _pressed()_ isn't obvious enough that you don't have to refer to the documentation or look at it's code. The same is equally true for _getInput()_ although renaming it to _buttonState()_ would help.

MLXXXp:
> You didn't go far enough to see if you put a heavily commented section at the end of the switch to explain this mess to anyone reading it

Here it is. However, if the author actually intended the original code to handle more than one button pressed at the same time, instead of it just happening to work, then I would stick with the original code using _if ()_s instead of a switch. Note that I haven't included the boundary tests that I mentioned previously.

``` cpp
// The switch will deal with checking if a button is pressed
// and set a new x or y location to move our text accordingly.
//
// If more than one button is pressed at a time, the first one
// found in the list will be handled and the rest will be ignored.

  switch (arduboy.buttonState()) {

    // if the right button is pressed move 1 pixel to the right every frame
    case RIGHT_BUTTON:
      x++;
      break;

    // if the left button is pressed move 1 pixel to the left every frame
    case LEFT_BUTTON:
      x--;
      break;

    // if the up button or A button is pressed move 1 pixel up every frame
    case UP_BUTTON:
    case A_BUTTON:
      y--;
      break;

    // if the down button or B button is pressed move 1 pixel down every frame
    case DOWN_BUTTON:
    case B_BUTTON:
      y++;
      break;
  }
```

joshgoebel:
> Which kind of proves my point. I guessed at how pressed() works because the intent wasn't fully expressed. The operation of pressed() isn't obvious enough that you don't have to refer to the documentation or look at it's code.

How would you rename it?  I think it's clear if you don't read into it.. exclusivelyPressed is closer to what you were thinking of.

> If more than one button is pressed at a time, the first one found in the list will be handled and the rest will be ignored.

Nope, If more than one is pressed NONE of them will match and your case will fail and not do anything. If you think it works (and you've tried it) it's only because you're not fast enough to hit both buttons at once - you're pressing one of them before the other.  Hold two down and your sketch should stop working because none of the statements will trigger.

MLXXXp:
> How would you rename it?

I wouldn't. I'm just saying that if I got it wrong, others might make the same mistake if they don't check the documentation or code (which is true for a good many other functions as well).

> Nope, If more than one is pressed NONE of them will match and your case will fail and not do anything.

Oops, you're right again. :cry: 
Change:

``` cpp
// If more than one button is pressed at a time, the first one
// found in the list will be handled and the rest will be ignored.
```

to:

``` cpp
// Only one button can be pressed at a time.
// If more than one button is held down at a time, all will be ignored.
```

And again, if that's not the desired operation, then I concede that using _if ()_ statements is likely a better choice. It would be up to the original author.

MLXXXp:
@yyyc514,
After thinking about this overnight, I'm going to eat a piece of humble pie and admit that I'm wrong and you are correct.

I still feel using _getInput()_ with a _switch/case_ statement is readable, suitable and efficient when you only want to handle single, individual button presses.

However, it's likely that many of the games and other types of sketches developed for the Arduboy will want to handle actions that result from more than one button being held down at a time. As you've pointed out, coding for these situations using a _switch_ could be difficult and confusing for less experienced programmers. 

Since this is the type of person  likely to be referring to _Buttons.ino_ in this PR as an example, it's much  more appropriate to use _if (pressed(...))_ for button handling, as is currently implemented.

I'm sorry to have wasted your time with this. I acknowledge your right to gloat, if you wish. :smile:

P.S. I still feel that boundary/out-of-range checks should be added before _Buttons.ino_ is accepted as an example.

joshgoebel:
@dragula96 Do you want to clean up the one or two things mentioned so we can get this merged or should we close it?

shogerr:
lol Mlxxxp, good post. We could just fork this and finish the PR to get it in there for the major release. I think Dragula got the thing going, and we want to tune it up, so might as well go ahead and do it?

joshgoebel:
You don't really "fork" it though... you'd just check out his branch, clean it up or commit on top of it, and then make a new PR.

Lets please squash this to a  single commit though if we're going to fix it ourselves.

MLXXXp:
Cleaned up code has been submitted as PR #87

PR #32, opened 2015-06-19
Arduboy core class (that other classes build upon)

joshgoebel:
This is a common pattern (seen in Gamebuino and other libs).  The problem is how to have very different "modes" (buffer, greyscale, tiled, etc.) but a single core lib.  Or even to answer the question of some people who want a terribly slim library vs those that want a LOT more functionality.  The core library can support both ways of thinking.  The answer is create a logical class hierarchy (not even a deep one).  I'd like to propose something like this:

(Print is included below to allow printing to screen, which comes from the Print.cpp Arduino core library)
- ArduboyCore (low-level hardware boot, button polling, power saving, safe mode, etc)
- Arduboy (ArduboyCore, Print, Buffered graphics) - same as today
- ArduboyText (ArduboyCore, Print, buffered text array)
- ArduboyTiled (ArduboyCore, support for array of tiles and buffered scrollable display rendered with tiles)

So pretty much just the core and then the classes MOST people would be using lay in on top of the core to provide the drawing modes, etc.  I'm not even saying all these classes need to be in core... it could be that ArduboyText and ArduboyTiled are developed as separate libraries and maintained in their own right.  Splitting the core hardware stuff into ArduboyCore allows people who want to build these libraries to still take advantage of the core lib, even if they're wanting to do something far different than a buffer backed display.  They still get the same boot up code, power saving, etc...

This encourages people to use the core lib (even if only for ArduboyCore) vs forking it into 5 different variants for different types of displays that do not want a buffer (or want a differently configured buffer).  

This pull request is the beginning of this effort.  It starts by creating ArduboyCore and moving all the hardware level stuff there.  Arduboy becomes a subclass of ArduboyCore and has almost exactly the same API as before.  I change `drawScreen` to `paintScreen` for clarify, but I could add a proxy `drawScreen` back to the `Arduboy` class if that would be helpful.

### DIFF
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -1,6 +1,46 @@
 #include "Arduboy.h"
 #include "glcdfont.c"
 
+const uint8_t PROGMEM lcdBootProgram[] = {
+  0xAE,  // Display Off
+  0XD5,  // Set Display Clock Divisor v
+  0xF0,  //   0x80 is default
+  0xA8,  // Set Multiplex Ratio v
+  0x3F,
+  0xD3,  // Set Display Offset v
+  0x00,
+  0x40,  // Set Start Line (0)
+  0x8D,  // Charge Pump Setting v
+  0x14,  //   Enable
+  // running this next pair twice?
+  0x20,  // Set Memory Mode v
+  0x00,  //   Horizontal Addressing
+  0xA1,  // Set Segment Re-map (A0) | (b0001)
+  0xC8,  // Set COM Output Scan Direction
+  0xDA,  // Set COM Pins v
+  0x12,
+  0x81,  // Set Contrast v
+  0xCF,
+  0xD9,  // Set Precharge
+  0xF1,
+  0xDB,  // Set VCom Detect
+  0x40,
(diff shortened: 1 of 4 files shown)


joshgoebel:
This also makes it easy to say reuse things like Adafruit_GFX without having to completely re-invent the wheel (unless we really, really want to).  Unless we're going to hand tune ALL those functions, why not stand on the shoulders of those who have come before.  Of course I'd optimize drawPixel and drawBitmap as much as you could even if you subclasses GFX, but mostly you could leave all the other methods alone and then our core library is a lot slimmer than before.  

It would look like this:

```
class ArduboyGFX: public Adafruit_GFX
{
void drawPixel(int16_t x, int16_t y, uint16_t color);

private:
unsigned char screenBuffer[(HEIGHT*WIDTH)/8];
}
```

Just implement drawPixel (that talks to screenBuffer) and then you're using Arduboy with the Adafruit_GFX library.  If they add a really cool graphics function, you get it for free - we don't have to worry about porting it over to Arduboy lib.

joshgoebel:
Ping.

PR #40, opened 2015-11-02
Fixed an issue where the ArduBreakout.ino example

tzikis:
There was an unprintable characted in 3 lines, which broke compilation (at least on Unix systems)

### DIFF
--- a/examples/ArduBreakout/ArduBreakout.ino
+++ b/examples/ArduBreakout/ArduBreakout.ino
@@ -367,7 +367,7 @@ boolean pollFireButton(int n)
   for(int i = 0; i < n; i++)
   {
     delay(15);
-    pad = arduboy.pressed(A_BUTTON) || arduboy.pressed(B_BUTTON);
+    pad = arduboy.pressed(A_BUTTON) || arduboy.pressed(B_BUTTON);
     if(pad == 1 && oldpad == 0)
     {
       oldpad3 = 1; //Forces pad loop 3 to run once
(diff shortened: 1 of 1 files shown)


shogerr:
Thanks, I'm thinking someone put a carriage return \r or something in there because it's sort of a long line.

shogerr:
Don't be davidperrnoud. We really need an automated build system in for both windows and linux that does a build status.

tzikis:
sorry guys, looks like i forgot to fix one of the errors, i think there's still one on line 311 (same place as in the other lines)

@rogosher well once the library is up on codebender, we take care of all that, that's actually how we found out about this issue :)

shogerr:
The main repository still needs a method to check PRs from users. Not sure what the solution is there, probably travis-ci or something. One of these days.

tzikis:
time for my shameless plug here :) we (codebender) doesn't have anything like that, but what you said yesterday actually got me thinking of building such a service exactly, and how we can leverage our existing infrastructure to build something like that

if you pay for that, we'll build it :)

also, please do fix the issue on line 311, i don't think it's big enough for me to make a separate PR :)

shogerr:
Removed the offending character. I am hoping checking in the unix like environment will be as simple as an automated makefile being ran. That's my hope anyway, or something like that. Not sure about windows atm.

tzikis:
great! now the problem should be solved!

well, the problem is a bit more complex, especially since the arduino ide doesn't make it too easy to call it in a makefile, and then there are things like different board configs (not applicable to arduboy but still valid), and potential dependencies with libraries (and most usually, specific versions of libraries)

then there is the issue of different arduino ide version. not all people convert to the latest as soon as it comes up. most people are still using 1.0.x, and I remember the Sprinter firmware was compatible with the horrible Arduino 0.22 for 2 years after 1.0.0 was released

PR #43, opened 2015-12-08
Add vertical and horizontal display flip functions

MLXXXp:
Adds **flipVertical()** and **flipHorizontal()** user functions to the core library. These send the appropriate commands to the SSD1306 display controller to flip (mirror) the contents of the display vertically or horizontally. If both are used, the display contents will be upside down.

If the boolean parameter passed to the function is _true_, the display will be set to flipped. If _false_, the display will be set to normal. After one of these function is used, the display will remain in that state until changed again, or the display is reset.

Are these useful for the Arduboy? They could be used to set the 0, 0 origin of pixel functions to any corner (and affected X/Y directions). Of course, any text output and bitmaps will be flipped accordingly.

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -211,6 +211,32 @@ void ArduboyCore::blank()
     SPI.transfer(0x00);
 }
 
+// flip the display vertically or set to normal
+void ArduboyCore::flipVertical(boolean flip)
+{
+  LCDCommandMode();
+  if (flip) {
+    SPI.transfer(0xC0); // reversed COM scan direction
+  }
+  else {
+    SPI.transfer(0xC8); // normal COM scan direction
+  }
+  LCDDataMode();
+}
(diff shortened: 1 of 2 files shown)


shogerr:
They are useful perhaps, but not without consulting the changes in the private repos for how its been done with Tetris. The developer working on Tetris <s>has no idea how to</s> doesn't use FOSS <s>effectively</s> so we won't be seeing that code anytime soon. It's kind of annoying, I'll leave this PR open for now and try and get some more info on what's been implemented so far and if the libs for Tetris are going to be merged with the master repo. If I get blank looks about merging that code into the master repo, it's probably a good idea to just go ahead and merge this, so it will be : )

shogerr:
Thanks for documenting everything. Looks great.

PR #44, opened 2015-12-18
optimised fillScreen function

spacebruce:
Should be way faster now, no need to do overdraw like in the breakout example.


spacebruce:
Oops, pulled wrong version, still learning Git, my apologies.

PR #45, opened 2015-12-18
Optimised fillScreen function

spacebruce:
Could possibly go even faster with bithax, but this is already much faster than using fillrect as there is no need for bounds checking. This should eliminate the need to do fiddly overdrawing such as in the breakout example, which will help with code clarity.

Works by setting every bit in every byte to the input color, doesn't require any code changes and only slightly increases library size.
If color is 0, B00000000
if color is 1, B11111111

### DIFF
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -359,6 +359,10 @@ void Arduboy::fillRect
 
 void Arduboy::fillScreen(uint8_t color)
 {
+  // C version : 
+  //if(color != 0) color = 0xFF;  //change any nonzero argument to b11111111 and insert into screen array.
+  //for(int16_t i=0; i<1024; i++)  { sBuffer[i] = color; }  //sBuffer = (128*64) = 8192/8 = 1024 bytes. 
+  
   asm volatile
   (
     // load color value into r27
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -383,9 +383,12 @@ void Arduboy::fillRect
 
 void Arduboy::fillScreen(uint8_t color)
 {
-  for(int16_t i=0; i<1024; i++) //1024 = (128*64)/8
+  color *= 0xFF;
   {
-    sBuffer[i] = color*256;     //if c=0, b00000000. if c=1, b=11111111
+    for(int16_t i=0; i<1024; i++) //1024 = (128*64)/8
+    {
+      sBuffer[i] = color;
+    }
   }
 }
 
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -383,7 +383,10 @@ void Arduboy::fillRect
 
 void Arduboy::fillScreen(uint8_t color)
 {
-  fillRect(0, 0, WIDTH, HEIGHT, color);
+  for(int16_t i=0; i<1024; i++) //1024 = (128*64)/8
+  {
+    sBuffer[i] = color*256;     //if c=0, b00000000. if c=1, b=11111111
+  }
 }
 
 void Arduboy::drawRoundRect
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -383,13 +383,35 @@ void Arduboy::fillRect
 
 void Arduboy::fillScreen(uint8_t color)
 {
-  color *= 0xFF;
-  {
-    for(int16_t i=0; i<1024; i++) //1024 = (128*64)/8
-    {
-      sBuffer[i] = color;
-    }
-  }
+  asm volatile
+  (
+    // load color value into r27
+    "mov r27, %1 \n\t"
+    // if value is zero, skip assigning to 0xff
+    "cpse r27, __zero_reg__ \n\t"
+    "ldi r27, 0xff \n\t"
+    // load sBuffer pointer into Z
+    "movw  r30, %0\n\t"
+    // counter = 0
+    "clr __tmp_reg__ \n\t"
+    "loopto:   \n\t"
+    // (4x) push zero into screen buffer,
+    // then increment buffer position
+    "st Z+, r27 \n\t"
+    "st Z+, r27 \n\t"
+    "st Z+, r27 \n\t"
+    "st Z+, r27 \n\t"
+    // increase counter
+    "inc __tmp_reg__ \n\t"
+    // repeat for 256 loops
+    // (until counter rolls over back to 0)
+    "brne loopto \n\t"
+    // input: sBuffer, color
+    // modified: Z (r30, r31), r27
+    :
+    : "r" (sBuffer), "r" (color)
+    : "r30", "r31", "r27"
+  );
 }
 
 void Arduboy::drawRoundRect
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -105,31 +105,7 @@ uint16_t Arduboy::rawADC(byte adc_bits)
 
 void Arduboy::clearDisplay()
 {
-  // C version:
-  // for (int a = 0; a < (HEIGHT*WIDTH)/8; a++) sBuffer[a] = 0x00;
-
-  // This implimentation should be close to an order of magnitude faster
-  asm volatile(
-    // load sBuffer pointer into Z
-    "movw  r30, %0\n\t"
-    // counter = 0
-    "eor __tmp_reg__, __tmp_reg__ \n\t"
-    "loop:   \n\t"
-    // (4x) push zero into screen buffer,
-    // then increment buffer position
-    "st Z+, __zero_reg__ \n\t"
-    "st Z+, __zero_reg__ \n\t"
-    "st Z+, __zero_reg__ \n\t"
-    "st Z+, __zero_reg__ \n\t"
-    // increase counter
-    "inc __tmp_reg__ \n\t"
-    // repeat for 256 loops
-    // (until counter rolls over back to 0)
-    "brne loop \n\t"
-    // input: sBuffer
-    // modified: Z (r30, r31)
-    : : "r" (sBuffer) : "r30","r31"
-  );
+  this->fillScreen(0);
 }
 
 void Arduboy::drawPixel(int x, int y, uint8_t color)


ace-dent:
Hi. For a more optimal solution, please look at the PR by @yyyc514 using ASM. By unrolling the loop, you can produce smaller and faster code:
https://github.com/yyyc514/ArduboyLib/commit/583e57476c468a23316ceec555ff8b0db0e39cda
Perhaps the FillScreen routine could at least fallback to the ClearScreen code where color is 0 (unset)...?

For new functionality, it might be interesting to add predefined dithered /checker board patterns that could 'fast fill' the display.
Color 3: 
1. B10101010
2. B01010101
Color 4: 
1. B01010101
2. B10101010
etc.

spacebruce:
Never had to do anything with ASM before, but this seems to do the trick. 
I'm not going to expand to patterns as that's outside the scope of what I was aiming for, but should be easy enough to add in

Pharap:
@MLXXXp If that's better then why isn't `clearDisplay` already doing this? And have you run any tests to prove `memset` works better or generates better code?

MLXXXp:
@Pharap, Perhaps because  @yyyc514 didn't consider investigating or using _memset()_? I'm just speculating, I don't know the history.

No, I haven't done any evaluation or testing but generally the functions in a language's standard library are well optimised.

Personally, I try to avoid using assembler unless it's absolutely necessary. It usually makes the code harder to understand and maintain, and requres more work to port the library to a processor with a different instruction set (such as an ARM based system, that Kevin B. is considering for Arduboy 2).

How often is _fillScreen()_ or _clearDisplay()_ used in a typical sketch? What percentage of overall time would be saved even if assembler proves to be slightly faster than _memset()_?

Pharap:
@MLXXXp 
Personally I think the example code is perfectly easy to read. Assembly isn't anything to be afraid of when working with embedded systems.

It's not too far off what memset does, except memset uses more jumps and uses `movw` more:

```
#if !defined(__DOXYGEN__)

#include "macros.inc"

#define dest_hi r25
#define dest_lo r24
#define src_hi r23
#define src_lo r22
#define len_hi r21
#define len_lo r20

    ASSEMBLY_CLIB_SECTION
    .global _U(memcpy)
    .type   _U(memcpy), @function
_U(memcpy):
    X_movw  ZL, src_lo
    X_movw  XL, dest_lo
#if OPTIMIZE_SPEED
; 15 words, (14 + len * 6 - (len & 1)) cycles
    sbrs    len_lo, 0
    rjmp    .L_memcpy_start
    rjmp    .L_memcpy_odd
.L_memcpy_loop:
    ld  __tmp_reg__, Z+
    st  X+, __tmp_reg__
.L_memcpy_odd:
    ld  __tmp_reg__, Z+
    st  X+, __tmp_reg__
.L_memcpy_start:
    subi    len_lo, lo8(2)
    sbci    len_hi, hi8(2)
#else
; 11 words, (13 + len * 8) cycles
    rjmp    .L_memcpy_start
.L_memcpy_loop:
    ld  __tmp_reg__, Z+
    st  X+, __tmp_reg__
.L_memcpy_start:
    subi    len_lo, lo8(1)
    sbci    len_hi, hi8(1)
#endif
    brcc    .L_memcpy_loop
; return dest (unchanged)
    ret
.L_memcpy_end:
    .size   _U(memcpy), .L_memcpy_end - _U(memcpy)

#endif /* not __DOXYGEN__ */
```

To put things into perspective though, I've done a breakdown of cycles.

To clarify, the screen buffer is 1024 bytes long ((128*64)/8).
Using memset, assuming the 15-word optimised version is being used, the cycle count is (14 + len \* 6 - (len & 1)). For 1024 bytes that's 5134 cycles.
For the assembly version in this pull request, the cycle count is (5 + (256 \* 9) + (255 \* 2) + 1), which is 2820 cycles.

This means memset is 2314 cycles more expensive, not accounting for the extra logic required for variable passing and function calling or the use of the ternary operator prior to the call to memset.

I don't have a device to hand to test this, but I think it's clear which is faster.
As to whether it's worth it or not, it depends entirely on how many times the user is going to be clearing the screen. I would assume most people would be doing it once at the start of a frame, which may not seem like much, but that's 2314 extra cycles the user then has available for their drawing logic, which could make all the difference in more graphics heavy games.

As for being portable, I think people expecting their Arduboy games to work exactly the same on an AVR Arduboy and a hypothetical ARM Arduboy II is being a bit optimistic. For a start systems like ARM, MIPS and PowerPC are much more complex in terms of memory mapping, caching etc.
Aside from which as long as the outward facing API remains the same then the implementation is free to be different.

(Personally I think they should have used ARM from the start because it's used by real-world consoles like the GBA, DS, 3DS and the PS Vita, as well as various tablets, the Raspberry Pi, and various phones. Plus Thumb is a very handy tecnhnology for limited memory systems.)

joshgoebel:
This matters a lot more for "painting" the screen from the buffer than clear or fill. When someone is trying to write an ambitious game at 60 FPS the amount of time render spends just moving the buffer to hardware becomes hugely significantly. 

If you want a portable slow library just use the adafruit stuff. 

Thanks,
Josh

shogerr:
+1 OK now we need a Chris.

Pharap:
@rogosher: you mean @arduboychris?

joshgoebel:
`memset()` increases compile size.  In order to make it worthwhile you need to start using it everywhere in order to pay for the complexity of the memset code.

joshgoebel:
I honestly don't think this should be assembly as a simple for loop in C is more than fast enough to achieve even probably 150 FPS without issue.  I think assembly should be used in a few places where speed is critical, and that's it.  Otherwise it just makes things harder to read for new coders and harder to contribute to.

Or else we split into a special assembly library where everyone can show off their assembly skills while leaving the core library mostly C.

Assembly should really only be used for:
- drawBitmaps (draw images to the buffer)
- paintScreen (move data from buffer to hardware)

You could make the argument to use it for complex draw operations too (lines, circles, etc) but god I have no interest at all in maintaining that code. :-)  I'm tempted to add font drawing, but honestly that's really just a higher-level variant of drawBitmaps (despite not being implemented that way now)... and also probably not speed critical... since after drawing the font you're probably stopping and giving the user a chance to read it... 

I suppose if you wanted to scroll a screen full of text at 60FPS this might start to matter a bit more.

joshgoebel:
:-1:  for reasons cited above regarding when we should and shouldn't drop to assembler.

Pharap:
I completely agree about memset being overkill unless used elsewhere, but I disagree about some of the other points.

For a start I think speed is always critical with all drawing functions considering that drawing has to be done on the cpu. The core library should try to be as fast and small as possible in order to give the end users of the code (i.e. the people writing games for the arduboy) more leeway in what they can achieve. When it comes down to it, this code base is the library for a commercial product, and as such the end users should be entitled to good quality code.

Assembly isn't that hard to read when done properly and used for small/simple things such as simple loops. As long as nobody is writing massive jumps or doing obscure things without giving reasonable justification and explanation in comments then assembly isn't all that hard to follow.

Besides which, if we avoid using advanced techniques just because the 'new coders' wouldn't understand the code when such techniques might be a superior solution then we're failing the end user.

This isn't exactly a function that's going to be modified a lot at any rate.

Lines and circles might benefit from being done in assembly, but you can't say either way until someone writes some code and provides some plausible explanation as to why it's quicker and precisely how much quicker it is (like the explanation given in this pull request).

I disagree about font not being speed critical. Drawing the score text on the hud of an action game takes up a lot more processing power than you'd expect.

Also, being pedantic, the core library is C++, not C. They are two different languages.

spacebruce:
So earlier it was 'C++ isn't good enough, use assembly', now people are saying the complete opposite. Can we have a consensus please?

Personally, I'd rather have a fast and efficient library over an easy to read one, every bit of speed is a good thing in the end, the programmer can make faster/more interesting games and players have a more fluid experience.
Besides, most users aren't going to be fiddling around changing library functions, and the ones that are able to make worthwhile changes shouldn't have any trouble working things out.

Pharap:
Currently the `clearScreen` function is implemented in assembly.
The code in this PR is effectively the same but adds the functionality of being able to set the (monochrome) screen to clear to either black or white instead of just black.
Which is to say it's effectively adding new functionality to an existing function.

Can someone please clarify why nobody wants to merge or close this PR?
Is it because it's in assembly? (Despite a similar function already existing)
Is it because it's a feature nobody wants? (Despite the fact it's cheap to add)

Personally I think it's useful and makes sense, practically all modern graphics libraries have the ability to clear the screen to a chosen colour (e.g. `glClearColour` & `glClear`).

joshgoebel:
> Personally I think it's useful and makes sense, practically all modern graphics libraries have the ability to clear the screen to a chosen colour (e.g. glClearColour & glClear).

We already have fillScreen for that... this is clear. Of course that is just an API naming question.  I feel like `clearScreen` and `fillScreen` are different things though despite.  And I've already said I think assembly here is overkill.  In an ambitious app all the clock cycles are lost moving screen screen from RAM to LCD, not in clearing between cycles. A for loop is more than fast enough here.

RE: Graphics if someone want stop re-write all the drawing functions in assembly be my guest, but it's not an interest of mine.  Most games for these platforms seem to make abundant use of sprites, so that's there I've spent my time over in my ArduboyExtra library - crazy fast sprites.

And with all things lets make sure we compare the compiled size before and after.

joshgoebel:
Oh I see now it's moving the functionality then reusing it.  Shrugs.  I guess that makes me a little less likely to object... 

Please don't remove the commented out C versions though, those make it easier for someone to hack on the library or learn.

Pharap:
@yyyc514 

There shouldn't be much of a size change at all. The modified assembly code for `fillScreen` only has a few more operations (about 2-3 I think, roughly 6 bytes) than the old version of `clearScreen`. `clearScreen` is then reimplemented as nothing but a call to `fillScreen` with a compile time literal argument, meaning that since the function is not virtual the compiler can simply inline the call to `clearScreen`, meaning that its use is equivalent to the user manually calling `fillScreen(0)`.

To put things into perspective, `clearScreen` is actually a special case of `fillScreen` where the fill colour is black (0) rather than white (1). In the current version of the library, `clearScreen` is implemented in assembly. This modification is taking the assembly code out of `clearScreen`, changing it to react to an argument so it can fill the screen with either black or white, placing it in `fillScreen` and then making `clearScreen` call `fillScreen`.

To argue that assembly is not required here is to argue that it was not required in `clearScreen` in the first place.

joshgoebel:
> To argue that assembly is not required here is to argue that it was not required in clearScreen in the first place.

It wasn't.  I was just bored. :-) _laughs_

Add comments C back and LGTM.

shogerr:
Alrighty @Pharap, please go ahead and put in the original C/C++ code, commented out, at the top of the function codeblock. After that, this PR will be merged. 

I think the case has been made, and objections heard. The end user will largely be calling the API from the context of their own sketches. If they would like to look under the hood, both the C/C++ and ASM versions will be available for clarity. This is a platform that is about learning after all : )

joshgoebel:
I squashed this into a single commit and we're merging it:
https://github.com/Arduboy/Arduboy/pull/65

joshgoebel:
Oh, future conversions to assembly need to come with benchmarks to show we're getting a real win for the huge additional complexity of converting to assembly.  10x win = good.  10% faster, uh maybe, maybe not, depending on the mess.

MLXXXp:
> Oh, future conversions to assembly need to come with benchmarks to show we're getting a real win

Agreed. To me, a main reason to avoid assembly is due to the extra difficulty involved with porting to a different CPU architecture.

Pharap:
Let's not forget, the assembly used here was a modification of code that already existed in the library to add functionality, it was not an entirely new addition. The code was originally submitted as C++ code, the function was changed to use assembly because of a suggestion by @ace-dent pointing out that the pre-existing `clearScreen` function used it. One assumes that if assembly was already being used in the library for a similar purpose that the argument as to whether is should actually be done in assembly has already been had. If the new function shouldn't have been done in assembly, then the original `clearScreen` function shouldn't have been done in assembly in the first place.

> a main reason to avoid assembly is due to the extra difficulty involved with porting to a different CPU architecture.

If it's possible to reimpliment this in C++ in a way that achieves comparable speed then by all means feel free to do so.

joshgoebel:
> One assumes that if assembly was already being used in the library for a similar purpose that the argument as to whether is should actually be done in assembly has already been had.

Not exactly similar purpose.  I should have added more comments perhaps.  clear/display is a HEAVY part of the render workflow and that makes it special... doing TONS of work (pushing bytes) and running 30, 60, 100 times per second.  It's one of the few things (other than perhaps image render) that SHOULD be crazy fast.  One shouldn't typically be running fillScreen 60 times a second.  The render pipeline should get it's job done as fast as possible so your application can either run it's application code or go back to sleep.

The only methods of equal importance (for most apps I've seen) are drawBitmap style functions that are used for frame rending and that could be drawn thousands of times per frame (and also push large amounts of data and do large amounts of math).  Technically this could apply to all the draw\* methods, but most of them are WAY too slow now to encourage their use in ambitious games (plus images are just cooler).  I dunno that we shouldn't accept assembly language versions of them, but I would tend to discourage it because I don't think those versions would be maintained - and I don't think people would be around to fix bugs when they were found.

It's true some other functions are frequently called, but many are super lightweight.  Button polling, for example, is likely done at 60 FPS also, but if button polling takes 1ns vs 10ns (order of magnitude) that doesn't affect anything.  It's just too tiny a sliver of time either way.  But if the screen render cycle goes from 25% CPU to 15% CPU that is a HUGE win that can result in faster gameplay or much improved battery life.

We just need to make sure we're speeding up the things that matter.  I'll go back and add comments regarding our usage of assembly.

Pharap:
> Not exactly similar purpose.

So `clearScreen()` isn't functionally equivalent to `fillScreen(Black)` or `fillScreen(White)`?
And users aren't going to want to use the `fillScreen` function as a means of choosing to fill (i.e. clear) the screen with a colour of their choice?

MLXXXp:
If you're constantly clearing the screen to white on each frame, and then writing black objects, you should probably consider using `invert()`.

joshgoebel:
> And users aren't going to want to use the fillScreen function as a means of choosing to fill (i.e. clear) the screen with a colour of their choice?

Probably not 60 times a second like the clear() function is used.  I never said it isn't functionally equivalent, just that it's not used in the same context.

> If you're constantly clearing the screen to white on each frame, and then writing black objects, you should probably consider using invert().

I see your point, but it won't make any difference... you're still filling the screen every frame with white or black. Also, just to frame this the expensive part of frames is moving the buffer from RAM to OLED, not clearing... it doesn't hurt to make clearing as fast as possible since we lose so much time on painting the OLED, but the law of diminishing returns is already in effect for clear.

MLXXXp:
> I see your point, but it won't make any difference... you're still filling the screen every frame with white or black.

I was addressing the comment @Pharap made:

> And users aren't going to want to use the fillScreen function as a means of choosing to fill (i.e. clear) the screen with a colour of their choice?

and I was implying that you should be using `invert()` along with `clear()`, avoiding `fillScreen()`.

joshgoebel:
Except you can't avoid it because clear is it. Either inverted or not you still (typically) call clear/fillScreen every loop to give you a blank slate. So there is no advantage in inverted.

Pharap:
> If you're constantly clearing the screen to white on each frame, and then writing black objects, you should probably consider using invert()

If that had been mentioned at the beginning then this could have been closed much earlier and the library could have been left as it was.

joshgoebel:
Again, inverted is the exact same process as normal just you always are drawing the opposite color.  Not sure how it helps unless what you truly want in an inverted screen.

PR #46, opened 2015-12-21
New Invert() and allPixelsOn() display functions

MLXXXp:
Added functions _invert()_ and _allOn()_ to the core library.

The _invert()_ function will command the display controller to invert the display. A pixel set to 1 will be off and vice versa. If the argument is _true_, the display will be inverted. If the argument is _false_ the display will be set to normal. An inverted display remains in effect until set back to normal or the display controller is reset.

Possible uses for this function:
- If having a white background with black objects is desired, it may be easier to work with an inverted display.
- Can be used to create effects such as explosions, lightning or a visual alert by briefly inverting the display and then setting it back to normal once or a few times.
- If what is being displayed is mostly a dark background with white objects, inverting the display during development of a program can help aid in determining if the location of an object is correct because the pixels between the objects become visible and the edge of the screen is easy to see.

The _allOn()_ function will command the display controller to set all pixels on, without affecting the contents of the display's RAM buffer. All pixels will turn on even if the display has been inverted using the above _invert()_ function. A _true_ argument will turn all pixels on and a _false_ argument will return to the previous display state.

Similar to the _invert()_ function, this _allOn()_ function could be used to simulate explosions or other flashes.

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -211,6 +211,32 @@ void ArduboyCore::blank()
     SPI.transfer(0x00);
 }
 
+// flip the display vertically or set to normal
+void ArduboyCore::flipVertical(boolean flip)
+{
+  LCDCommandMode();
+  if (flip) {
+    SPI.transfer(0xC0); // reversed COM scan direction
+  }
+  else {
+    SPI.transfer(0xC8); // normal COM scan direction
+  }
+  LCDDataMode();
+}
+
+// flip the display horizontally or set to normal
+void ArduboyCore::flipHorizontal(boolean flip)
+{
+  LCDCommandMode();
+  if (flip) {
+    SPI.transfer(0xA0); // reversed segment re-map
+  }
+  else {
+    SPI.transfer(0xA1); // normal segment re-map
+  }
+  LCDDataMode();
+}
+
 
 /* Buttons */
 
--- a/core.cpp
+++ b/core.cpp
@@ -211,6 +211,21 @@ void ArduboyCore::blank()
     SPI.transfer(0x00);
 }
 
+
+// invert the display or set to normal
+// when inverted, a pixel set to 0 will be on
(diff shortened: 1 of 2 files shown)


MLXXXp:
Please note:
I added these new functions after adding the _flipHorizontal()_ and _flipVertical()_ functions in my working repository. If this pull request is merged "as is" these flip functions, that I made pull request #43 for, will be included as well.

Sorry if this causes any inconvenience.

joshgoebel:
I'm not sure why "allOn" is even necessary, but code here looks good.

joshgoebel:
LGTM

MLXXXp:
@yyyc514 said:

> I'm not sure why "allOn" is even necessary, but code here looks good.

Like I said, quickly flashing the entire display on and then back to it's previous content, without affecting the processor's display buffer or the controller's refresh RAM, could be used to simulate flashes, hits, warnings, etc., with minimal code overhead.

Anyway, none of these functions will add any code to the sketch if they're not used.

I have no objection to renaming _allOn()_ to _allPixelsOn()_. It describes the purpose better.

joshgoebel:
Let's change name then and LGTM

Thanks,
Josh

MLXXXp:
Function name _allOn()_ changed to _allPixelsOn()_.

joshgoebel:
Should probably now be rebased on my code here:

https://github.com/Arduboy/Arduboy/pull/47

joshgoebel:
Can you update on latest master to use sendByte or whatever we named it and then I think we can get this merged.

MLXXXp:
My plan was to do it yesterday but I ran out of time. Today for sure.

MLXXXp:
A new pull request #50 has been created to implement the same functions using _sendLCDCommand()_.

I'm closing this pull request.

PR #47, opened 2016-01-30
abstract sending single byte commands

joshgoebel:
- saves 34 bytes compiled (when using both functions)

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -211,32 +211,30 @@ void ArduboyCore::blank()
     SPI.transfer(0x00);
 }
 
-// flip the display vertically or set to normal
-void ArduboyCore::flipVertical(boolean flip)
+void ArduboyCore::sendLCDByteCommand(uint8_t command)
 {
   LCDCommandMode();
-  if (flip) {
-    SPI.transfer(0xC0); // reversed COM scan direction
-  }
-  else {
-    SPI.transfer(0xC8); // normal COM scan direction
-  }
+  SPI.transfer(command);
   LCDDataMode();
 }
+
+#define OLED_VERTICAL_FLIPPED 0xC0 // reversed COM scan direction
+#define OLED_VERTICAL_NORMAL 0xC8 // normal COM scan direction
+
+// flip the display vertically or set to normal
+void ArduboyCore::flipVertical(boolean flipped)
+{
+  sendLCDByteCommand(flipped ? OLED_VERTICAL_FLIPPED : OLED_VERTICAL_NORMAL);
+}
+
+#define OLED_HORIZ_FLIPPED 0xA0 // reversed segment re-map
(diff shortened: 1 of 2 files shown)


MLXXXp:
I wouldn't name it _sendLCDByteCommand()_. Name it _sendLCDcommand()_.

This way, if required later, you don't need a _sendLCDtwoByteCommand()_, _sendLCDthreeByteCommand()_, etc.

You can just overload the function:

``` cpp
/// send a single byte command to the OLED
void sendLCDcommand(uint8_t command);
/// send a two byte command to the OLED
void sendLCDcommand(uint8_t command1, uint8_t command2);
// (etc. ...)
```

Also, I know I'm just being picky and it's already been used previously ( _LCDcommandMode(), LCDdataMode(), bootLCD()_ ), but using _LCD_ in function names that control the display is a bit confusing, since the display uses OLED technology, not LCD. I would change all instances of _LCD_ in the library to a more generic term _display_ or _screen_. This would also be of benefit if the library were to be ported to some other display technology.

joshgoebel:
> This way, if required later, you don't need a sendLCDtwoByteCommand(), sendLCDthreeByteCommand(), etc.

The intention was to distinguish between a word command in the future which could use a uint16t but otherwise the same signature call in the source, so if someone was just glancing at a single function call with no other context handy they would immediately now if it was a byte or word call.  We do have some word commands for the display.

Now the question you bring up is if those calls would be better expressed as passing of two bytes instead of a word, and you might be right there.

joshgoebel:
> If you're going to start defining names for the display commands then it would be nice if it were done for all of the ones used throughout the entire library. All these command #defines should be grouped in one location in a header file.

No objection.  The only other place so far is the LCD boot though and I'm not sure a ton of constants would add the same value there.

joshgoebel:
Changed name of function.

MLXXXp:
> Now the question you bring up is if those calls would be better expressed as passing of two bytes instead of a word, and you might be right there.

The SSD1306 display controller documentation describes all commands in terms of bytes. So I think passing two bytes as separate arguments is better.

The are also quite a few three byte commands and even some longer, which I think are also best sent as individual bytes. That is, if it's decided that any of these commands are useful.

The display controller doesn't really care when you switch between command and data mode. You can raise and lower D/C between each command byte and/or stay in command mode across multiple commands. Therefore, one possible version of overloading _sendLCDCommand()_ (if proven useful) could be an array pointer and a length:

``` cpp
void sendLCDCommand(uint8_t *commands, uint8_t length);
```

joshgoebel:
Ha, then you could pass lcdBootProgram directly to that. My gut says at this point it would use up a few extra bytes but if we ever start sending more than one long strict of bytes to the LCD it might make sense to wrap that up with a function.  Not the kind of thing I'd add until it was needed.

PR #48, opened 2016-01-30
store pin boot program in array

joshgoebel:
- saves 22 bytes when compiled

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -1,5 +1,21 @@
 #include "core.h"
 
+const uint8_t PROGMEM pinBootProgram[] = {
+  // buttons
+  PIN_LEFT_BUTTON, INPUT_PULLUP,
+  PIN_RIGHT_BUTTON, INPUT_PULLUP,
+  PIN_UP_BUTTON, INPUT_PULLUP,
+  PIN_DOWN_BUTTON, INPUT_PULLUP,
+  PIN_A_BUTTON, INPUT_PULLUP,
+  PIN_B_BUTTON, INPUT_PULLUP,
+
+  // OLED SPI
+  DC, OUTPUT,
+  CS, OUTPUT,
+  RST, OUTPUT,
+  0
+};
+
 const uint8_t PROGMEM lcdBootProgram[] = {
   0xAE,  // Display Off
   0XD5,  // Set Display Clock Divisor v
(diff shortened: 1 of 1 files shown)


MLXXXp:
I wonder how many pins would have to be programmed to make it more efficient to have two arrays; one for INPUT_PULLUP and one for OUTPUT?

The arrays would contain just the pin numbers, so would be half the size (plus the terminator). You would call _bootPins()_ twice and pass an array pointer and mode to it as arguments.

MLXXXp:
@yyyc514, A bit off topic for this pull request but if you're trying to save some program space taken up by the library, you could look into not sending initialisation commands to the display that just set the values defaulted to after a display reset.

I was able to eliminate the sending of quite a few commands in the [SparkFun MicroView library](https://github.com/geekammo/MicroView-Arduino-Library), which uses the same display controller (but with a smaller screen). See function _MicroView::begin()_ in _MicroView.cpp_ and look for the commented out commands as examples.

joshgoebel:
> I wonder how many pins would have to be programmed to make it more efficient to have two arrays; one for INPUT_PULLUP and one for OUTPUT?

Yeah some of it is effort vs guesswork vs my gut vs what looks cleanest.  You could also have a MODE, pin, pin, pin, pin, next_mode, pin pin pin, end... but then the logic to process it gets a bit larger... might win, might not, but this is an easy win and what we're doing is still fairly straightforward.  A nice win for a first pass.

I know about sizeof, was trying to avoid all that and let it self terminate.  Really?  What is pin 0?  I'm tempted to leave it or for correctness do you really think it needs to change?  As you said we aren't using pin 0... and it'll compile smaller with a 0 compare than a fixed value compare.

> A bit off topic for this pull request but if you're trying to save some program space taken up by the library, you could look into not sending initialisation commands to the display that just set the values defaulted to after a display reset.

Yes, that's another potential area for savings but I wasn't in a rush to go there because maybe some hardware is a little different or something?  Seems to be one area (more so than others) that could potentially trip up someone... most of my changes are "do the same thing, do it 10% smaller"...  but if you want to work on a PR for that it could also be a win.

I just find things as I see them.  I don't sit down JUST to make the library tiny.

joshgoebel:
There is a TON of fat in the audio stuff if you're looking for a place to seriously hack on... I just attack this little stuff because I see it and it's easy and hey, Rome wasn't built in a day. :)

MLXXXp:
> Really? What is pin 0? I'm tempted to leave it or for correctness do you really think it needs to change?

Arduino digital pin 0 (ATmega32U4 pin PD2) is the USART1 Rx pin for serial communication, so it's highly unlikely that it would be used as a general purpose input or output, but it can be. Therefore, keeping 0 as the terminator isn't really a problem, if changing it causes any efficiency issues or you're otherwise uncomfortable in doing so.

I mainly pointed it out because not being careful of what you use as a terminator can sometimes cause "gotcha"s with future additions to a list.

joshgoebel:
> I mainly pointed it out because not being careful of what you use as a terminator can sometimes cause "gotcha"s with future additions to a list.

Agree.  I was just assuming someone adding port 0 to the list would also change the terminator, but maybe that's hoping for too much. :)

MLXXXp:
>  I was just assuming someone adding port 0 to the list would also change the terminator

But if the pin number has already previously been abstracted by a _#define_, as all the ones in the array currently are, a person adding it may not notice that its value is 0, and then wonder why the change doesn't work. ;-)

Even using 0xFF as a terminator isn't totally safe, but it's unlikely that there will be a pin number 255. 

Anyway, if using 0xFF generates more code, I'd say leave it at 0.

PR #50, opened 2016-02-04
Add invert() and allPixelsOn() display functions

MLXXXp:
invert(true) will invert the display so a 0 bit in the buffer will
turn the corresponding pixel on. invert(false) will set to normal.

allPixelsOn(true) will set all pixels on, ignoring the contents of
the buffer but leaving it unaltered. allPixelsOn(false) will display
the contents of the buffer.

See the discussion in obsolete pull request #46 for more info.

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -231,6 +231,26 @@ void ArduboyCore::sendLCDCommand(uint8_t command)
   LCDDataMode();
 }
 
+#define OLED_PIXELS_INVERTED 0xA7 // All pixels inverted
+#define OLED_PIXELS_NORMAL 0xA6 // All pixels normal
+
+// invert the display or set to normal
+// when inverted, a pixel set to 0 will be on
+void ArduboyCore::invert(boolean inverse)
+{
+  sendLCDCommand(inverse ? OLED_PIXELS_INVERTED : OLED_PIXELS_NORMAL);
+}
+
+#define OLED_ALL_PIXELS_ON 0xA5 // all pixels on
+#define OLED_PIXELS_FROM_RAM 0xA4 // pixels mapped to display RAM contents
+
+// turn all display pixels on, ignoring buffer contents
+// or set to normal buffer display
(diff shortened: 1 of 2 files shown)


MLXXXp:
Here's a "quick and dirty" sketch I wrote to test the functions.

``` cpp
#include <Arduboy.h>

#define DELAY_TIME 2000

Arduboy aboy;

void setup() {
  aboy.start();
}

void loop() {
  aboy.clearDisplay();
  aboy.setCursor(0, 0);
  aboy.println("Normal");
  aboy.display();
  delay(DELAY_TIME);

  aboy.print("All On");
  aboy.display();
  aboy.allPixelsOn(true);
  delay(DELAY_TIME);
  aboy.println(" .. Normal");
  aboy.display();
  aboy.allPixelsOn(false);
  delay(DELAY_TIME);  

  aboy.println("Inverted");
  aboy.display();
  aboy.invert(true);
  delay(DELAY_TIME);

  aboy.print("All On");
  aboy.display();
  aboy.allPixelsOn(true);
  delay(DELAY_TIME);
  aboy.println(" .. Normal");
  aboy.display();
  aboy.allPixelsOn(false);
  delay(DELAY_TIME);  

  aboy.print("Invert off");
  aboy.display();
  aboy.invert(false);
  delay(DELAY_TIME);
}
```

joshgoebel:
Any reason not to throw that in the examples folder with a small readme?

MLXXXp:
Be my guest.

It might be nice to also suggest some practical uses in the readme. Ones I can think of:
- Either _invert()_ or _allPixelsOn()_ could be toggled once or twice quickly to simulate explosions, hits, lightning, etc.
- During debug, when there's a mostly black background, temporarily using _invert()_ can show the edge of the screen and allow counting pixels in the space between objects.
- For a sketch that would look better with dark objects on a white background, it may be easier to use _invert()_ once during init. A _clear()_ will clear to all white and objects will be drawn in black.

PR #52, opened 2016-02-04
Added keywords.txt file

MLXXXp:
I've added a _keywords.txt_ file for syntax highlighting of the Arduboy library functions, types, etc. when using the Arduino IDE.

I didn't include the audio stuff because I'm not sure how mature it is and I haven't researched what should be exposed to a sketch developer.

I might have also missed a few other functions, and there may be some that shouldn't be there, but it's a start.

I've put everything in alphabetical order, which IMHO makes it easy to maintain.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -1,0 +1,58 @@
+#######################################
+# Syntax Coloring Map For Arduboy
+#######################################
+
+#######################################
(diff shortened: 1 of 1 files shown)


shogerr:
Thank you : )

PR #53, opened 2016-02-05
Added allPixelsOn function to keywords.txt

MLXXXp:


shogerr:
I think this will add a better user experience. The keywords.txt file can be produced another way, but a template is a nice place to start.

Right now, making the library most use able with the Arduino IDE is the best approach. We can change that later though.

MLXXXp:
If _keywords.txt_ were auto-generated it might include keywords that a sketch developer probably shouldn't use, for example: ArduboyCore, ArduboyAudio, ArduboyTunes.

For the frequency that classes and functions will be added to the Arduboy library, I don't think manually maintaining *keywords.txt" should be a problem. It only took me about 10 minutes to create it from scratch.

joshgoebel:
> If keywords.txt were auto-generated it might include keywords that a sketch developer probably shouldn't use, for example: ArduboyCore, ArduboyAudio, ArduboyTunes.

There is no reason you can't use them directly though depending on what you're wanting to do, and having them in the file certainly wouldn't harm anything if they weren't used.  I think long-term we should look into making this an automatic process.

Is this for highlighting or also autocomplete? (I don't use the IDE but for compiling)

MLXXXp:
> Is this for highlighting or also autocomplete?

I'm pretty sure it's just highlighting. The [Keywords section of the the Arduino IDE 1.5 Library Specification](https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification#keywords) recommends that a _keywords.txt_ file be included.

PR #55, opened 2016-02-06
reduce compiled size of drawChar

joshgoebel:
- use a simple loop rather than linking in fillRect
- 208 bytes savings (sketch not using drawRect)

I didn't time it but this is quite likely faster as well since it
doesn't have the overhead of a stack of function calls resulting in tons
of jumps and push/pops to/from the stack.

### DIFF
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -579,6 +579,7 @@ void Arduboy::drawSlowXYBitmap(int16_t x, int16_t y, const uint8_t *bitmap, int1
 void Arduboy::drawChar
 (int16_t x, int16_t y, unsigned char c, uint8_t color, uint8_t bg, uint8_t size)
 {
+  boolean draw_background = bg != color;
 
   if ((x >= WIDTH) ||         // Clip right
     (y >= HEIGHT) ||        // Clip bottom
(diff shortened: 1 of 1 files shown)


shogerr:
Matching this PR to the existing formatting style will make it easier to merge this before a utility to do some automated refactoring is looked at. 

For the time being see: http://homepages.inf.ed.ac.uk/dts/pm/Papers/nasa-c-style.pdf

joshgoebel:
>  will make it easier to merge this

It's a clean merge already - I wouldn't submit a PR that wasn't.  When a proper tool and guidelines are agreed on then we can see figure out a process to make sure all the files confirm that that - currently the various files are a conglomeration of different styles.

I find the Linux kernel style much more rational:
https://www.kernel.org/doc/Documentation/CodingStyle

yyyc514:
I can link to coding style guidelines also: https://www.kernel.org/doc/Documentation/CodingStyle

;-)

joshgoebel:
Compile size is going to depend on your exact sketch.  The MORE use everything the less savings, the less you you everything the more expensive every single thing is (if they are all interconnected).

I was using the Music sketch which does itself use the large text feature.  So it was a direct comparison against a program using that function but not fillRect, drawFastVLine, etc... a program ALREADY using all those functions would show less savings (such as Breakout).  And this is even more important because even programs that DON'T draw text, or don't draw LARGE text still pay this penalty of having fillRect, drawFastVLine linked and compiled even if they never used at all.  See the issue thread on that problem... so most sketches (all the TeamARG stuff, etc.) still will save the 208 bytes, and for some it will save less.  

_But if speed is your goal you should be staying away from all the primitive drawing functions anyways except for drawImage._  All of them are varying degrees of slow to just ok compared to rendering sprites directly against the buffer.

I have no idea why it would be slower though - it's doing the same x/y looping that fillrect and drawFastVLine are but without the CALL/PUSH/POP penalty... it's possible the extra math could be the problem but again drawing large text isn't a high performance need for most people.

If you wanted to post an issue with your code and exact use case that might be interesting.

Issue #11, opened 2015-05-29
Add arduboy logo to boot sequence


joshgoebel:
![screen shot 2015-05-29 at 3 53 24 pm](https://cloud.githubusercontent.com/assets/6473/7891466/e8e44a36-061a-11e5-95d1-bfa71fcb82eb.png)

Can we agree that (by default) unless a sketch chooses not to compile it (for space reasons) that at boot time we should scroll in the Arduboy logo?  I downsized the graphics from Arduboy.com and rasterized it exactly but if there isn't agreement on this I probably won't go further.

I thought it would be nice to have some consistency and build in a standard "device experience" that would include toggling brightness, sound, boot logo, etc.  Kind of help with the branding of the Arduboy experience.

Another option would be to compile the boot sequence into the Bootloader so that it's completely outside of a sketch... I'm very curious how much space is to spare there.  I know the default on Atmega32U4 is to use 4k for the boot loaded, but I"m not sure how much of that space is actually used by the boot loaded now.

joshgoebel:
It looks pretty slick.  I have it worked over in https://github.com/yyyc514/ArduboyLib/tree/brightness if anyone wants to try it.  I'll make a sep PR later.  I did add a define so devs could easily turn it off if they really really didn't want it - though I honestly think it should be part of the hardware (bootloader).

joshgoebel:
Is this welcome in core?  If not I'm going to go ahead and put it in the Extra lib I'm building for Arduboy.  If it's welcome I'll make a PR.

shogerr:
Yes, please make a PR for this. The branding is very slick. I agree that if the developer wishes to override the boot logo the option should be easily available.

joshgoebel:
The difficulty is how to turn it off... it's the kind of thing you'd want to use defines for, but the root sketch file of a project has no way to set defines that are evaluated before the entire project is compiled... and if you make it a programatic option then in my experience C is still going to link it in the final sketch (even if it's never called)... because technically it could have been called.

That means whatever space we spent storing the logo is wasted for anyone not wishing to show the logo... unless they are comfortable editing the library itself, in which case there is no problem.  All the problems go away if we accept most people might not use the core library as-is but rather would download it and then edit one of the H files to set preferences, etc.

shogerr:
[...]

PR #57, opened 2016-02-07
Bootup Arduboy logo scroll-in and red power LED

joshgoebel:


joshgoebel:
The first attempt has my "fade in" attempt if anyone is curious about that.

joshgoebel:
This is about 416 bytes larger than without.   I wanted to use the hardware to do the scrolling to keep the cost of flash as low as possible.  904 bytes to use a drawImage loop and just do it the easy way (if the sketch isn't otherwise using the built in drawImage).  Extra 294 bytes if you were already using drawImage.  A lot of pro games are going to use faster or more specialized draw routines though so they will be paying the full 904 byte penalty.

Any feeling here on how high is too high?  3.7% of Flash (based on boatloader and "minimum" sketch size of Blink) just to get our logo on their seems to be paying a bit much.  I'd have been happy with 1.x percent which is about where my hardware based approach was going to land us - but I'm not sure it's viable.

I don't have memory usage for fade in off the top of my head but I think it was half as small.  Scrolling from the side would also be lightweight since we don't have any complex bit shifting to do, but then you lose the whole Gameboy feel.

The compile size of master now is (latest IDE) 12,554 (compiling built in Tunes example, which is plenty good enough for testing the boot logo).  So that's what we're working from.

joshgoebel:
Someone else feel free to poke around with this and see if you can get it smoother - and ask if you have any questions.  YOu'll need to check out this branch then checkout '8d89f759fce99e0e1a788551fc74032448d4495b' to get the hardware scroll code.

joshgoebel:
Actually my fade code is currently +588 bytes.

joshgoebel:
444 bytes but it still flickers a little.  Anyone else want to try it and see what they think?

Don't merge this till I clean it up, it's a mess - but it's testable.  I want to see if the flicker is annoying to anyone else.

246 bytes just for a static non-moving logo.

joshgoebel:
All sorts of cool effects you can make the OLED do by hacking the commands but none of them looked super intentional.

MLXXXp:
I don't recall if it's already been previously discussed, but have you considered just putting the static non-moving logo in the center of the screen, then pausing a bit, then using the hardware horizontal scroll command to scroll the logo left, like a marquee, for another short while?

I don't really have time right now to try it myself, but here's what to do:
- Display the logo
- Pause, using delay() or whatever, for a short time.
- Send a  _Deactivate Scroll_ command 0x2E, required as per section 10.2.1 in the [SSD1306 datasheet](http://www.buydisplay.com/download/ic/SSD1306.pdf).
- Send a _Horizontal Left Scroll_ command: 0x27, 0x00, _start page_, _speed_, _stop page_, 0x00, 0xFF. The _start page_ and _stop page_ should cover the logo height, but you could use 0x00 and 0x07 as a start, for the entire screen. You can fiddle with the _speed_ for what looks best, but I'd start with 0x07 for the fastest. See the datasheet for the _speed_ values, they're not linear or in order.
- Send an _Activate Scroll_ command: 0x2F.
- Pause again to show the scrolling for a while.
- Send a _Deactivate Scroll_ command: 0x2E.
- Clear the screen, unless it's done later, or just let the sketch do it.

If you wish, you could similarly experiment with the simultaneous horizontal and vertical (diagonal) scroll command: 0x29 or 0x2A.

You could also try right horizontal scrolling, using 0x26 instead of 0x27 but I'm guessing it wouldn't look as good as left scroll.

joshgoebel:
Feel free to do it and submit a PR for us to play with - but the interest has always been in a VERTICAL scroll, not horizontal.  I'm aware horizontal is easier (whether you ask the hardware to help or not.

It's too bad we don't have a vertical ONLY hardware scroll.

MLXXXp:
I spent a little time trying horizontal scroll. It works. The commands and delays past the first delay take 160 bytes but could probably be optimised like the main display init. I didn't include any sound.

I'm quickly writing this comment and don't have time to create a PR. You can just cut and paste the _Arduboy::bootLogo()_ function with:

``` cpp
void Arduboy::bootLogo()
{
  const unsigned char *logo = arduboy_logo;
  uint16_t mem_position = 20+128*3;
  uint8_t c=0;

  c = 0;
  while (c < 88*2) {
    sBuffer[mem_position++] = pgm_read_byte(logo);
    logo++;
    c++;
    if(c==88) {
      mem_position+=128-88;
    }
  }
  display();

  delay(800);

  sendLCDCommand(0x2E); // disable scroll

  sendLCDCommand(0x27); // left horizontal scroll
  sendLCDCommand(0x00); // dummy
  sendLCDCommand(0x00); // start page
  sendLCDCommand(0x07); // speed
  sendLCDCommand(0x07); // end page
  sendLCDCommand(0x00); // dummy
  sendLCDCommand(0xFF); // dummy

  sendLCDCommand(0x2F); // enable scroll

  delay(3650);

  sendLCDCommand(0x2E); // disable scroll

  delay (500);
}
```

joshgoebel:
What is the total byte difference from master?

Thanks,
Josh

MLXXXp:
> What is the total byte difference from master?

I don't know what method you're using to calculate this. Would you mind trying it out yourself? If you think it's something worth considering, then you could do the calculations. There could be differences between our development environments that might cause differing results.

MLXXXp:
@yyyc514, I did try your latest vertical scroll logo. I didn't find the flicker too annoying. My vertical scroll version is smoother though. :smile: 

Note that the homebuilt system I'm using is still wired as a DevKit equivalent and I only have a 0.96" screen. The flicker may be more noticeable on a 1.3" screen.

joshgoebel:
> I don't know what method you're using to calculate this. 

Check out latest master.  Compile a super simple sketch that doesn't really use any functions that you might be considering (I use Tunes.ino).  Write down the compiled size.  Check out your code that does boot animation.  Compile. Write down the size. Subtract.

That's the size you're adding.  The exact size would depend on 100s of variables, but for a ballpark of worse case this is the best way to get that number.

I'd think if we're on the same IDE version we should get exactly identical results.

> My vertical scroll version is smoother though.

If you did vertical would like to see the numbers.  It's super smooth if you just use the drawing functions but then you increase the size significantly for anyone not wanting to use those own functions themselves.

MLXXXp:
> > My vertical scroll version is smoother though.
> 
> If you did vertical would like to see the numbers.

Sorry, that was a brain fart. I meant "My _horizontal_ scroll version" (using hardware scroll).

I didn't start from the master branch, but what I did before posting the code for my version of the _Arduboy::bootLogo()_ function, was this:

With IDE 1.6.7 and the following sketch:

``` cpp
#include <Arduboy.h>

Arduboy arduboy;

void setup() {
  arduboy.begin();
}

void loop() {
}
```
- Grabbed your _yyyc514:hw_scroll_ branch as a zip file and set it up as my Arduboy library.
- In _core.h_ I commented out `#define ARDUBOY_10` and uncommented `#define AB_DEVKIT` to work with my hardware. 
- Compiled it as-is (and loaded it to see how the logo looked). Size: 9348.

Everything following is changes to function _Arduboy::bootLogo()_
- Deleted the two lines of speaker related code at the end of your version, since I wasn't going to have sound in my version. Size: 9344
- Changed the function to become my version above. Size 9264
- Deleted all the code in my version past the `delay(800);` to the end, the part that does the scrolling. Size: 9104
- Deleted the `delay(800);` so only the logo display code and a `display()` remained, to show just a static logo. Size 9092
- Deleted everything. leaving just `void Arduboy::bootLogo() { }` which generates no logo code at all. Size 8802

Since I used your code to generate the logo, and only changed `uint16_t mem_position = 20+128*5;` to `uint16_t mem_position = 20+128*3;` to change the location. Your version is also size 9092 if everything past the first `display()` is removed.

### Conclusion:

Your version with the speaker lines removed is 9348 - 8802 = 546
My version is 9264 - 8802 = 462
So my version is 546 - 462 = 84 bytes smaller than yours without sound.

Both versions have identical initial logo display code: 9092 - 8802 = 290

Your scrolling and delay code is 9348 - 9092 =  256
My scrolling and delay code is 9264 - 9092 = 172

MLXXXp:
I've now replaced the 9 consecutive `sendLCDCommand()` functions with an array sent by a `for` loop. The  total is now 9242, so all the logo code is 9242 - 8802 = 440 bytes. This is an additional savings of  22 bytes.

Here's the new code:

``` cpp
const uint8_t PROGMEM HorizScrollCmd[] = {
  0x2E, // disable scroll

  0x27, // left horizontal scroll
  0x00, // dummy
  0x02, // start page
  0x07, // speed
  0x05, // end page
  0x00, // dummy
  0xFF, // dummy

  0x2F  // enable scroll
};

void Arduboy::bootLogo()
{
  const unsigned char *logo = arduboy_logo;
  uint16_t mem_position = 20+128*3;
  uint8_t c=0;

  c = 0;
  while (c < 88*2) {
    sBuffer[mem_position++] = pgm_read_byte(logo);
    logo++;
    c++;
    if(c==88) {
      mem_position+=128-88;
    }
  }
  display();

  delay(800);

  LCDCommandMode();
  for (int8_t i=0; i < sizeof(HorizScrollCmd); i++) {
    SPI.transfer(pgm_read_byte(HorizScrollCmd + i));
  }
  LCDDataMode();

  delay(3650);

  sendLCDCommand(0x2E); // disable scroll

  delay (500);
}
```

joshgoebel:
Can you push this to a branch of your repository?

MLXXXp:
> Can you push this to a branch of your repository?

[https://github.com/MLXXXp/Arduboy/tree/hw_scroll](https://github.com/MLXXXp/Arduboy/tree/hw_scroll)

MLXXXp:
I modified my test sketch to add calls to _clear()_, _display()_ and _delay()_. The compile size is now 9302 with my logo code, and 8918 with all logo code removed.

So, given that a sketch itself uses both _display()_ and _delay()_, the impact of including the logo is 9202 - 8918 = 384 bytes.

I also tried it with _delay()_ removed from the sketch. There was no change, still 384 bytes (It turns out to be because  _ArduboyCore::bootPins()_ in _core.cpp_ also uses _delay()_).

New test sketch:

``` cpp
#include <Arduboy.h>

Arduboy arduboy;

void setup() {
  arduboy.begin();
  arduboy.clear();
  arduboy.display();
//  delay(1000);
}

void loop() {
}
```

joshgoebel:
It's smoother, but it needs to slide in, not just slide around.   I think we need to find a way to just use our own drawBitmap and do the animation right.  The smooth movement is just so sweet.  Really ambitious games using a faster drawBitmap could just modify the library to use their own (getting the space back).

Thoughts?

MLXXXp:
> It's smoother, but it needs to slide in, not just slide around.

**needs** to slide in? Is that just your opinion or has it been mandated somehow? It does something more than just a static display. It's kind of different and unique and I don't find it particularly annoying or anything.

joshgoebel:
I find it cheap to be honest.  It looks like a screen saver, not an intro.  No offense.  Kevin (the creator of Arduboy) liked it sliding in from the top because it has that retro Nintendo feel to it.

MLXXXp:
So just an opinion. (No offence taken). Has anyone else seen it and commented?

joshgoebel:
I'm going to push an update that also uses the RED led on power on... it's starting to feel "right", at least to me. :-)

MLXXXp:
By @yyyc514 in incident #11:

> All we need is begin() and beginWithoutLogo() if we want to make the logo super easy to remove on a per-sketch basis...

Any thoughts on still doing this? You implied that it wouldn't have any impact on code size since only one would be complied at a time.

joshgoebel:
> Any thoughts on still doing this? You implied that it wouldn't have any impact on code size since only one would be complied at a time.

It wouldn't.  Still entirely doable, just not sure on the naming.  We could still get in the boot tone either way if someone has audio on since that should be small. :-)

Thanks for reminding me of that, kind of lets us forget about the sizing a little more.

shogerr:
Can we overload the `Arduboy` constructor in the source file and set a default parameter so it's off; avoiding a new function all together, but keeping the `Arduboy.h` file as is? Then a user could just choose to pass a variable to the constructor turning off the boot sequence? Also, did we ever think about using an 8 bit data type to store 8 flags for the system? Might be worth it get some efficient booleans going for system settings.

MLXXXp:
> Can we overload the Arduboy constructor in the source file and set a default parameter so it's off; avoiding a new function all together, but keeping the Arduboy.h file as is?

I don't think doing anything with the constructor would help, but if you overload _begin()_ with a dummy argument, I'm pretty sure it would work. E.g.

``` cpp
#define NO_LOGO 1
void begin();
void begin(uint8_t dummy); // dummy is not actually used. its value doesn't matter
```

A sketch using `arduboy.begin()` would include the logo code.
A sketch using `arduboy.begin(NO_LOGO)` would not include the logo code.

The code for the `begin(uint8_t dummy)` version would ignore the parameter. It would have to be written so the logo is suppressed non-programmatically.

However, this is almost identical to having two separate functions, e.g. `begin()` and `beginWithoutIntro()`.

I don't think the overloading technique is any more comprehensible than having two functions. Actually, the fact that the parameter is just a dummy, to force different code to be compiled under the covers, may be confusing to a sketch developer.

Also, `begin(NO_LOGO)` might result in slightly more code than `beginWithoutIntro()` because `begin(NO_LOGO)` would require some additional code to pass the argument. :wink:

joshgoebel:
> Can we overload the Arduboy constructor in the source file and set a default parameter so it's off; avoiding a new function all together, but keeping the Arduboy.h file as is?

Off or on, I though we wanted it to be on by default?

> Also, begin(NO_LOGO) might result in slightly more code than beginWithoutIntro() because begin(NO_LOGO) would require some additional code to pass the argument.

Oh god, are you thinking like me now? :-)  Honestly though the compiler _should_ be smart enough to figure out you never use the variable and not generate any code for dealing with it/passing it.

I purposely picked a somewhat ugly name to discourage people from turning it off for no reason.  If you're sketch is big and you legit need to turn it off it's not hard to just look in the header and copy the function name once and be done with it.  I thought it was a nifty idea.

Either way there have to be two copies of the function for the compiler - so was it the two copies you were choking on, or just the naming?

MLXXXp:
> Either way there have to be two copies of the function for the compiler - so was it the two copies you were choking on, or just the naming?

Are you asking me or @rogosher? My preference is your original suggestion; two separate functions:
`begin()` to display the logo
`beginWithoutLogo()` (or some other meaningful name) to leave the logo code out.

I'm sure you know this, but for the possible benefit of others reading this, you don't need two copies with lots of code, that differ only in one not having the logo. It's more maintainable to make two additional functions containing common code; `beginCodeFirstPart()` and `beginCodeLastPart()` (or whatever, for the names). Then your two functions are:

``` cpp
void Arduboy::begin() {
  beginCodeFirstPart();
  bootLogo();
  beginCodeLastPart();
}

void Arduboy::beginWithoutLogo() {
  beginCodeFirstPart();
  beginCodeLastPart();
}
```

joshgoebel:
> beginCodeFirstPart() and beginCodeLastPart()

Ugh... and we're pretty much already doing that already.  The function already is just a small dispatch center to other functions.  You could also do something crazy with macros, but it's easier and clearer to just have two short functions and comment the difference between them.

If the routine was 50 lines long instead of 5 I might have considered splitting it up, but only because it should have been... just just not pre and post functions for the heck of it.

shogerr:
Overloading the constructor would just look cleaner. 

```
Arduboy arduboy(); //normal, include boot
\\or
Arduboy arduboy(0); //no logo
```

Where

```
\\constructor
Arduboy::arduboy(int logo = 1) { ... }
```

Or some other reasonable init process.

joshgoebel:
> `Arduboy::arduboy(int logo = 1) { ... }`

You realize that then you're just cutting and pasting the same code into two different constructors, right?  Plus then if someone uses an int (vs a DEFINE, which you can't force) you have no idea reading the code what it means.

`begin()` and `beginWithoutLogo()` are clear and impossible to confuse.  It's super clear what is going on.  I don't understand what is "clean" about your proposal.

shogerr:
if you don't pass a value to the overloaded construct it will use the paramater supplied in the overload definition, in this case 1, so no those calls are not the same. 

<strike>I have a hard time seeing such long hungarian notation in a c/c++ program is one issue : P But we've adopted it so it's all good.</strike> See issue #83

joshgoebel:
I didn't say they were the same.  I said you need TWO constructors if you want to compiler to be able to exclude the boot function.  You can't pass a variable and do `if` later because then the compiler will compile in the boot sequence no matter what because at _compile time_ it doesn't know whether you will pass a 1 or 0 in your code - so it keeps ALL the code just in case.

Also you may not want to do your initialization in your constructor... doing it your way would force running the whole boot inside the constructor since you use any conditional logic or save state for later.

shogerr:
Yes! now thats a comment that helps eliminate the suggestion. Good work on this so far, the led stuff is a nice touch. I think I should have a production unit in a bit and can properly test it out myself.

joshgoebel:
>  I think I should have a production unit in a bit and can properly test it out myself.

Or change it to use the BLUE color of the RGB and try it on the devkit. :-)

MLXXXp:
> ... just just not pre and post functions for the heck of it

It's for maintainability. If someone not intimately familiar with the code decides to add some new code to `begin()`, they might not realize that it also has to be copied and pasted into `beginWithoutLogo()`, So now you've got one function with the new code and one without. If the code was to fix a bug, then the bug hasn't been fixed in `beginWithoutLogo()`

Also, If you've made changes in two different functions, you have to test them both to make sure a mistake wasn't made, like pasting into the wrong spot, or missing a section when you copied, or a typo if you just manually entered the code twice. If there's only a single change that will be used by both funcitions then your testing requirements can be relaxed.

Thirdly, there's extra time and effort involved (no matter how small) in making the changes in two spots instead of one.

And what if we decide to add a third _begin()_  function, such as `beginWithStaticLogo()` to reduce the code footprint somewhat but still display some branding? This would only differ from `begin()` in that it would call `bootStaticLogo()` in place of `bootLogo()`. And how about a fourth: `beginWithoutLogoAndFlashlight()` to free up even more space than `beginWithoutLogo()`. Now identical changes have to be made in three or four places, and you have to test each one of them.

joshgoebel:
> It's for maintainability. If someone not intimately familiar with the code decides to add some new code to begin(), they might not realize that it also has to be copied and pasted into beginWithoutLogo()

That's why we have comments and peer review.

> Also, If you've made changes in two different functions, you have to test them both to make sure a mistake wasn't made, like pasting into the wrong spot,

Again, peer review + as long as this is mostly just a list of function calls then C does most of the actual checking for you at compile time.

> And what if we decide to add a third begin() function, such as beginWithStaticLogo() to reduce the code footprint somewhat but still display some branding? 

If it turns out this is common then we've picked the wrong solution and we need to just tell people to do their own boot sequence manually how ever they want, and not call begin().  But for logo or no logo I think it will be fine.  It's not the beginning of a trend.

MLXXXp:
> That's why we have comments and peer review.

So more effort and the chance that the review missed something.

Other than a few more bytes used to make the two calls, what's the downside of functions for common start and end sections?

joshgoebel:
> Other than a few more bytes used to make the two calls, what's the downside of functions for common start and end sections?

Functions should be named clearly.   Functionals should read clearly.  If you have an actual patch with REAL function names that is still readable and doesn't look like it was ONLY done to appease the compiler monsters then lets see it, but it's not `beginCodeFirstPart();` and friends.

But **again**, the function is already concise as it is with minimal code before and after the call.  There is not a problem here that needs a solution. 

I have 2 functions that read clearly now, at a glance.  Splitting them into 4 that is even more of a jumble is not an improvement.  You'll have to trust me on this one.

joshgoebel:
**Eventualy** the audio lines are begging to be consolidated into a sep function, but I'm still waiting for us to figure out what we're doing with tunes before caring too much about it - and I would do it regardless of this discussion regarding logo - because it's audio related and makes it's own argument to be grouped... if it was a few lines longer it probably would have been broken out already.

But the reason to break it out is (of course) for clarity, not just to have a "before" and "after" stubs to please the compiler so you can copy and paste a little less.

That said, there is also some logic for leaving this as an easily copyable list... if someone DOES want to tweak the boot sequence further the best way to do it is to go to begin() and just copy the method and then make their own changes.  So if this had ALL 15 (eventual) bootup steps wrapped in a single (or double) method, that wouldn't be so bad...

I finally think you're just making the argument too early... if we start to add more and more code to both methods (if we get to 10 or 15 line of boot "modules") then I'll think of a better solution... but at only 5-6 lines, they are fine.

One could imagine:

```
// copy and paste this instead of using Arduboy::begin if you want to make your own customized
// start up
begin()
{
bootCore();
bootUtils();
bootLogo();
bootRAM();
bootSD();
bootAudio();
}
```

Just an example.  That's probably an exaggeration since if something truly became part of the hardware spec it would likely end up in the core itself and be covered in the low-level hardware init. But you get the idea.

This could also be a compiler define and part of the board file if we were to encourage it's use.  Though then that puts the decision more into the hands of the compiler vs sketch creator.

MLXXXp:
> Functions should be named clearly. Functionals should read clearly.

That's why we have comments

If you need a comment that says:
"// If you have to make changes in either begin() or beginWithoutLogo() you likely have to change both because they each consist mostly of identical code."

how is that any better than comments that say:
"// This function is to wrap code that is common to both begin() and beginWithoutLogo() to so that changes are only required in one place to affect both of them."

joshgoebel:
I maintain the function(s) are clear as they are.  Sorry if you do not agree.  If this becomes an issue in the future we'll look into other options.

Arduboy.cpp
```diff
@@ -715,3 +773,4 @@ void Arduboy::swap(int16_t& a, int16_t& b)
   a = b;
   b = temp;
 }
+
```
shogerr: Missing newline at end of file

ab_logo.h
```diff
@@ -0,0 +1,28 @@
...
+0xC0, 0xE0, 0x70, 0x3F, 0x1F, 0x00, 0x00, 0x1F, 0x3F, 0x70, 
+0xE0, 0xC0, 0xC0, 0xC0, 0xE0, 0x70, 0x3F, 0x1F, 0x00, 0x00, 
+0x7F, 0xFF, 0xC1, 0xC1, 0xC1, 0xC1, 0xC1, 0xE3, 0x77, 0x3E, 
+0x1C, 0x00, 0x00, 0x1F, 0x3F, 0x70, 0xE0, 0xC0, 0xC0, 0xC0, 
+0xE0, 0x70, 0x3F, 0x1F, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 
+0xFF, 0xFF, 0x01, 0x00, 0x00, 0x00
+
+};
+
+#endif
```
shogerr: Missing newline. See, http://stackoverflow.com/questions/72271/no-newline-at-end-of-file-compiler-warning

The c++ standard isn't linkable anywhere online I don't think.

joshgoebel: You're saying these are compiler warnings for you?  I'll look again.

shogerr: Missing newlines will cause compiler warnings if the compiler implements the c++ standard fully. Newlines at the end of files is just good practice, and would need to be in files merged, is more so what I'm getting at.

MLXXXp:
> I maintain the function(s) are clear as they are. Sorry if you do not agree. If this becomes an issue in the future we'll look into other options.

Sure, no problem. Let's just agree to disagree :smile:

Since there doesn't appear to be enough "peer reviewers" voicing their preferences, to indicate the more desirable approach, we could argue back and forth until we're blue in the face.

ab_logo.h
joshgoebel: I haven't seen this cause real problems in practice other than the weird looking diff from time to time, and I've never seen a merge trip up on the end of a file because of the missing newline, but maybe sometimes it does. _shrugs_ (i'm referring here to end-of-file newlines)

I added new lines though and changed my editor setting (my editor isn't set up as well on this laptop as my work machine).

shogerr:
I want to make one final proposal before merging this PR.
Instead of `beginWithoutArduboyBootLogo()`, does `beginNoLogo()` sound reasonable?

joshgoebel:
I can rename it.

I was really hoping someone would spend some time coming up with a really nice sound... did we decide this one was "good enough".  I wasn't trying to pick the boot sound for Arduboy, just show an example of a fun effect.

Or should we merge this without sound until we have a REAL sound?

shogerr:
Let's put in a sound. Just make it an unobtrusive something, something that says, "Hi! I'm on.". That will do for this merge, this way we can get this merged and open an issue for the sound design.

joshgoebel:
Removing sound.  We shouldn't end up with a sound just because I did something and no one else had to time to actually give it any real thought.  It's fine for this to be silent.  If their game has music as soon as it loads they will know their sound is working anyways.

People who want to play with the sound should look at the new PR I just created. (it's based on this branch but it can be rebased when we have a final solution)

PR #58, opened 2016-02-07
small flash memory savings add up (50 bytes)

joshgoebel:
- saves 10 bytes
- two variables do not need default values
- cheaper to call setFrameRate and let it assign defaults

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -17,43 +17,37 @@ const uint8_t PROGMEM pinBootProgram[] = {
 };
 
 const uint8_t PROGMEM lcdBootProgram[] = {
-  0xAE,  // Display Off
-  0XD5,  // Set Display Clock Divisor v
-  0xF0,  //   0x80 is default
-  0xA8,  // Set Multiplex Ratio v
-  0x3F,
-  0xD3,  // Set Display Offset v
-  0x00,
-  0x40,  // Set Start Line (0)
-  0x8D,  // Charge Pump Setting v
-  0x14,  //   Enable
-  // running this next pair twice?
-  0x20,  // Set Memory Mode v
-  0x00,  //   Horizontal Addressing
-  0xA1,  // Set Segment Re-map (A0) | (b0001)
-  0xC8,  // Set COM Output Scan Direction
-  0xDA,  // Set COM Pins v
-  0x12,
-  0x81,  // Set Contrast v
-  0xCF,
-  0xD9,  // Set Precharge
-  0xF1,
-  0xDB,  // Set VCom Detect
-  0x40,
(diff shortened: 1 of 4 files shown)


joshgoebel:
Ready for review.

core.cpp
```diff
@@ -17,43 +17,42 @@ const uint8_t PROGMEM pinBootProgram[] = {
...
-  0xAF,  // Display On
-
+  // boot defaults are commented out but left here incase they
+  // might prove useful for reference
+  //
+  // Further reading: https://www.adafruit.com/datasheets/SSD1306.pdf
+  //
+  // 0xAE,     // Display Off
+  0XD5,     // Set Display Clock Divisor v
+    0xF0,   //   0x80 is default
```
shogerr: Seems to be some whitespacing issue here and below. If these are reminders of some sort, they aren't naturally intuitive, so maybe we match all the white space or explicitly state what the offset indicates?

shogerr:
There are some interesting white spacing issues in `core.cpp`, that's all I see off hand.

core.cpp
MLXXXp: The indents indicate the second byte of a two byte command.

shogerr: So could we maybe separate 2-tuples by new lines then? Or rather, every command is padded by at least one newline on the top and bottom? Might make things easier to read. 

That would create some length to the file, but because these are pretty ambiguous to someone browsing through the source, it certainly doesn't hurt to be verbose about what command does what, so each command could be preceded by a comment defining it, the bytes below, and the second byte with an inline comment appended perhaps?

joshgoebel: If you want to spend the time reworking this I"d do:

```
// Set Display Clock Divisor v; 0x80 is default
0xD5, 0xF0,
// Next command sequence; second comment
0x00, 
// etc
```

I honestly liked the spacing and since the next step was commenting out most of them and just leaving them for reference I didn't think it worth the time to rewrite the whole section. If someone else wants to do that in a PR, ok.

joshgoebel:
> There are some interesting white spacing issues in core.cpp, that's all I see off hand.

True that, but I'm one to fix things as they come up (I love git blame and actual revision history).  I have a branch with a Makefile and some source cleanup tools laying around - when i get around to working on that we'll reformat all the files at the same time and discuss it in a PR.

Right now I just do the work and try not to make the formatting worse.

joshgoebel:
So shall we merge this and then someone cleans up the command sequence IF they want?

This is the one blocker for 0.9.

joshgoebel:
Good now?

shogerr:
The formatting you put in looks great, thanks for that.

PR #67, opened 2016-02-09
Arduino library style guide compliance

MLXXXp:
Changes are mainly to be more compliant with the [Arduino Style Guide for Writing Libraries](https://www.arduino.cc/en/Reference/APIStyleGuide).

Note: Example sketch file _Tunes.ino_ was changed from DOS CR/LF line termination to Unix LF termination (and _arduboy.start()_ changed to _arduboy.begin()_).


Arduboy.cpp
```diff
@@ -682,14 +687,12 @@ unsigned char* Arduboy::getBuffer()
 
 boolean Arduboy::pressed(uint8_t buttons)
 {
- uint8_t button_state = getInput();
- return (button_state & buttons) == buttons;
```
joshgoebel, line 686: Did you compare compile size for this change? If it's not smaller I like the clarity of the previous code.

Arduboy.h
```diff
@@ -40,12 +40,13 @@ class Arduboy : public Print, public ArduboyCore
...
-   * if (not_pressed(LEFT_BUTTON))
+   * if (notPressed(LEFT_BUTTON))
    */
-  boolean not_pressed(uint8_t buttons);
+  boolean notPressed(uint8_t buttons);
 
   /// Initializes the display.
-  void start();
+  void begin();
+  void start() __attribute__ ((deprecated("use begin() instead")));
```
joshgoebel, line 46: Perfect.

Arduboy.cpp
MLXXXp, line 686: The change doesn't seem to affect compile size.

Personally, I find my version clearer. I tend to avoid defining local variables that are only used once, unless they're being assigned by some long, complicated expression. But, we can change it back, unless other people have an opinion.

joshgoebel, line 686: Please change it back.  I wrote it that way on purpose.  Though honestly my desire for it probably reflects on the poorly named getInput() more than anything else (because I don't disagree with your logic).

What about `pollButtonState()`?

joshgoebel:
I already fixed the Music sketch in master.  Make sure you already start from the latest master when making a PR.  Do you know how to rebase?

Arduboy.cpp
joshgoebel, line 686: >  I tend to avoid defining local variables that are only used once, 

Just so you're aware though there isn't any type of inefficiency in this since the compiler knows it's only used once - the compiled code probably is exactly the same - since the compiler needs a variable/register to put the result in anyways whether you ask for one or not.

MLXXXp:
> Please change it back.

Done. It's your code so you're the boss. :smile:
But I changed the indenting from one space to two, to match the rest of the formatting, if that's all right with you.

> Make sure you already start from the latest master when making a PR.

It was the lastest master when I started working on it and there were no conflicts when I submitted the PR.

> Do you know how to rebase?

I'll figure it out.

Arduboy.cpp
MLXXXp, line 686: > What about pollButtonState()?

At this point, I think _getInput()_ has been used too heavily to justify changing it. I don't think _getInput()_ is too confusing, as long as it's well documented (Doxygen makes that easy, doesn't it?)

joshgoebel:
> Done. It's your code so you're the boss. 

I was open to renaming the function also... I just found clarity in it being clear that what you were working with was the button state.  Probably will go ahead and submit a PR renaming it. :-)

MLXXXp:
Hopefully I've got the conflicts resolved properly and this PR can be merged.

MLXXXp:
> Probably will go ahead and submit a PR renaming it.

Remember to change the example sketches to match. And we should strive to come up with a good way to inform sketch developers that the old name has been deprecated, so they can update existing sketches, and more importantly, to make sure new sketches use the new name instead of using old unchanged sketches as an example and just copy/pasting.

shogerr:
@MLXXXp I think for the 1.0.0 release I should probably write out a nice long changelog, or include a wiki for developers? I don't think a community page would be suited to inform here, and sometimes changelogs don't get looked at. This seems a good use of the github wiki. That way the migration info will always be attached to the github repo.

I'll be able to take a look at this PR again tonight properly as well, that is in a 4 or 5 hours from this post.

joshgoebel:
> And we should strive to come up with a good way to inform sketch developers that the old name has been deprecated, so they can update existing sketches, and more importantly, to make sure new sketches use the new name instead of using old unchanged sketches as an example and just copy/pasting.

Changelog + deprecation notices?

joshgoebel:
Not sure what you tried, but it wasn't the right thing.  I rebased and merged here:

https://github.com/Arduboy/Arduboy/pull/77

PR #72, opened 2016-02-10
declare ArduboyCore static

joshgoebel:
- one core to rule them all!
- also fixes inline compile warnings


joshgoebel:
I need to test this on the real hardware but it should work and fixes the inline warnings.

MLXXXp:
Is there a reason `setRGBled()` isn't declared static, like all the others?

joshgoebel:
Omission.

PR #73, opened 2016-02-10
uint8_t for width, height, radius arguments

joshgoebel:
- 44 bytes saves for Ardubreakout example sketch
- if you're using images/shapes taller/wider than 255 you
  made need to rethinking what you are doing

### DIFF
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -163,7 +163,7 @@ uint8_t Arduboy::getPixel(uint8_t x, uint8_t y)
   return (sBuffer[(row*WIDTH) + x] & _BV(bit_position)) >> bit_position;
 }
 
-void Arduboy::drawCircle(int16_t x0, int16_t y0, int16_t r, uint8_t color)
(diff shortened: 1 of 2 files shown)


MLXXXp:
I think the only reason for allowing >255 (and it's an unlikely one) would be if future enhancements to the functions allowed objects to be placed partially off screen with the out-of-bounds portions clipped. E.g. a circle centered at 260, 260 with radius 150.

Even so, if support for this was made available, we could always change everything back to int16_t. It probably wouldn't affect most sketches, if any, due to implicit integer promotion.

joshgoebel:
> I think the only reason for allowing >255 (and it's an unlikely one) would be if future enhancements to the functions allowed objects to be placed partially off screen with the out-of-bounds portions clipped. E.g. a circle centered at 260, 260 with radius 150.

This is already allowed.  These are the width, height and radius values we're talking about.  We currently (AFTER this change) allow objects 2-4x larger than the entire screen... that seems entirely unnecessary.

joshgoebel:
Ok sketches needing to do this could copy the methods and make them ints just for their own use. But drawing things that large will also be slow.

PR #75, opened 2016-02-10
reduce storage of note frequences (-32 bytes)

joshgoebel:


audio.cpp
```diff
@@ -117,11 +120,16 @@ void ArduboyTunes::playNote(byte chan, byte note)
...
 
   // we only have frequencies for 128 notes
-  if (note > 127)
-    note = 127;
+  if (note > 127) {
+    return;
+  }
 
   timer_num = pgm_read_byte(tune_pin_to_timer_PGM + chan);
-  frequency2 = pgm_read_word(_midi_note_frequencies + note);
```
flaki, line 124: Ha! This is clever!

(Although I do wonder how the added code increases size. Do we have numbers on the objdump before/after? @rogosher)

joshgoebel, line 124: See the byte count in the commit notes.  That is aggregate.  This is a win.

flaki, line 124: Ah all right, sorry! Yeah it was pretty clear this was a win, I was just wondering how big of a win, thanks for the heads-up!

shogerr:
@flaki Thanks for asking the right questions : ) The rule about questions applies here I think, if you have a question, then someone else is the community probably has the same question, so just ask it!

audio.cpp
joshgoebel, line 124: That's why I starting putting the savings in the commit messages, so it's clear and no one would have to ask. :)

PR #77, opened 2016-02-11
standardize a bit more with other Arduino Libraries

joshgoebel:
Rebase and cleaned up Scott's great work on this.

### DIFF
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -30,7 +30,7 @@ void Arduboy::begin()
   // Audio
   tunes.initChannel(PIN_SPEAKER_1);
   tunes.initChannel(PIN_SPEAKER_2);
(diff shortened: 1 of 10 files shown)


joshgoebel:
What you want to read about is interactive rebasing, once you learn how you'll never go back.

I just checkout out your branch, rewound your last merge of itself (no idea what's with that) then rebased interactively onto currently master... snipped your last 2 "fix notPressed more" from the end and snapped them right after " Change not_pressed() to notPressed()" and then told git to make them all part of the same commit... so now it looks like you just did the work once, and all part of a single commit.

Makes for really super clean commits once you learn the pattern.

joshgoebel:
Oh I guess I didn't fix the commit text about removing the variable, but that's also possible.

Issue #15, opened 2015-06-04
Publish library to Arduino Library Manager


joshgoebel:
Reference:

https://github.com/arduino/Arduino/wiki/Library-Manager-FAQ

Looks like just adding the metadata (my PR already does this) and then just asking they add it to their batch job... and tagging official releases (so their job finds them).

shogerr:
Yes, thank you. Please add these files. At the very least, the link is extremely helpful.

joshgoebel:
We need to hurry towards a 1.0 and then tag something to get this moving on their end.  Or at least maybe tag a 0.8 or 0.9?

shogerr:
Let's go with 0.9

joshgoebel:
I'll tag what we have now as 0.8.5 (it fits plus it's nostalgic from when I started Ruby on Rails).  Plan to tag 0.9 in the next day or two.  We should go ahead and reach out to Arduino to get added to the official list though.

shogerr:
Alright, I'm caught up. I've created a draft for the v1.0.0 release and would like to use that method instead of the tag method unless there is a reason not to. 

We are pretty much ready for the release too. It would be nice to get the boot sequence done for v1.0.0

Also, I haven't had a chance to look through their library specification guide, https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification so I'm not sure how well we fall in line with that. Should probably do an audit before submitting the issue.

MLXXXp:
One main thing to decide is whether the library should be Arduino 1.0.x backward compatible or should be 1.5.x only. This will determine the directory structure.

One drawback to making it 1.0.x compatible is that when people use _Sketch -> Include Library_ in the IDE, it will pull in additional unnecessary headers other than _Arduboy.h_, such as _audio.h_ and _assets.h_. If 1.5.x only, this could be avoided by putting the headers in subdirectories.

Conversely, the advantage of 1.0.x compatibility is self explanatory.

joshgoebel:
> If 1.5.x only, this could be avoided by putting the headers in subdirectories.

Can you confirm that actually works?  I haven't found subdirectories to work.  Link to example lib doing that?

MLXXXp:
> I haven't found subdirectories to work.

You have to use relative paths for any _#include_s for files not in the same directory.

For a directory structure such as:

```
├── examples
│   └── Tunes
│       ├── README.md
│       └── Tunes.ino
├── extras
[...]

PR #79, opened 2016-02-12
break out printing into separate class

joshgoebel:


joshgoebel:
We should discuss the right way to do this, no real way to depreciate these items and leave them because that defeats the purpose (to rip them out so C++ removes them from the build).  Should this drop after 1.0 and just be a breaking change for any program using text rendering?

I mean we could add some stubs for depreciated calls, but I'm honestly not sure most devs (or not devs) are looking at the compile logs for compiler warnings to see what they need to fix.  I think that is the 1% of us.

MLXXXp:
I'm for it. There will probably be a good number of sketches that don't use these functions. As long a we provide good documentation for those that do want to use them, things should be fine.

And, I agree we should do it fast. Lots of people are going to start receiving production Arduboys very soon, and many will want to start writing code for it. Rookie programmers are likely to want to play with printing text to the screen (can you say "Hello World" :wink:)

joshgoebel:
@rogosher Once I test this should we put it in 1.0 to get it out "fast" (since it's breaking) or hold it till later (cause it's breaking)?

MLXXXp:
Nitpicking:

I'd like to see instances of _ab_ changed to _aboy_. I.e. _ab_printer.h_ to _aboy_printer.h_, _AbPrinter_ to _AboyPrinter_. That is, if you feel it's too long to use _Arduboy_ completely. The same goes for _ab_logo.h_, if we decide to include logo code.

It took me a second to determine what _ab_ stood for, but I would recognise the abbreviation _aboy_ quickly.

shogerr:
> @rogosher Once I test this should we put it in 1.0 to get it out "fast" (since it's breaking) or hold it till later (cause it's breaking)?

@yyyc514, Let's hold it, we have a few days anyhow.

I really like _aboy_ as our common abbreviation, same with `AboyPrinter` over `AbPrinter`. When reading code, you know instantly what _aboy_ stands for and it is still only 4 characters. Also, we do want to stick with camel casing per #83. I think that applies to file names? Yes/No?

joshgoebel:
I prefer lowercase (more compatible across filesystems for one) but it looks like Arduino seems to prefer camel-case.  There is no technical reason we need to switch though unless you insist.

**It's not an abbreviation, it's just a prefix to namespace the names.**   I'd prefer to not have it at all but the more classes we have the more important it becomes to namespace them and the full "Arduboy" (or even Aboy)  makes no sense for more generic glasses... eventually we should have some nice non-systems objects... like physics, and shapes... and it's NOT an AboySquare or ArduboyGravity... We're really talking about squares and gravity.. We only put "AB" in front of them to prevent them from conflicting with other Arduino libraries a user might have installed.  Does that make sense?

See Apple's own namespacing for big groups of things.. NSString, UIDynamicAnimator... 

Some classes may make sense to use the full name... like ArduboyCore, etc... and we can still do that for those things, but it shouldn't be a requirement for every class or every file.

MLXXXp:
> Also, we do want to stick with camel casing per #83. I think that applies to file names? Yes/No?

You would think so, but I think they only really care about function names (especially those exposed to the user of the library), not file names.

Near the end of the [Style Guide](https://www.arduino.cc/en/Reference/APIStyleGuide) they cite [https://github.com/adafruit/Adafruit-BMP085-Library](https://github.com/adafruit/Adafruit-BMP085-Library) as being an exemplary example of a library, and it uses underscores in the library name itself.

joshgoebel:
Oh. I agree here. I think the style guide cares more about the naming that programmers have to interact with constantly vs the names - which you put in a header once and never look at again.

shogerr:
For the namespaces, do we consider 4 chars to long? If no, then lets do _aboy_. If yes, then what is the thought on 3 chars? If that is ok, then how about _aby_? If neither 3 or 4 chars will work, then _ab_ it is.

MLXXXp:
Just a note related to my comments in issue #15 and the changes in PR #91:

Files _ab_printer.h_ and _ab_printer.cpp_ (as well as _glcdfont.c_) would **not** go into a subdirectory under _src/_, as I previously indicated. They would reside directly in the _src_ directory since they are part of the user API.

This would mean that, along with `#include Arduboy.h`,  a `#include ab_printer.h` would be added to the developer's sketch when _Sketch -> Include Library -> Arduboy_ was used in the IDE. This is what we want to happen, since _ab_printer.h_ would need to be included by any sketch using the _AbPrinter_ class.

Including _ab_printer.h_ wouldn't impact a sketch not using the _AbPrinter_, so the `#include ab_printer.h` could just be left in. If a developer didn't like the idea of the sketch including a header file that wasn't actually used, the line could just be removed manually. We could note this option in the _AbPrinter_ documentation to make it clearer.

MLXXXp:
> **It's not an abbreviation, it's just a prefix to namespace the names.** I'd prefer to not have it at all but the more classes we have the more important it becomes to namespace them and the full "Arduboy" (or even Aboy) makes no sense for more generic glasses... eventually we should have some nice non-systems objects... like physics, and shapes... and it's NOT an AboySquare or ArduboyGravity... We're really talking about squares and gravity.. We only put "AB" in front of them to prevent them from conflicting with other Arduino libraries a user might have installed.

If the object creation class/library is truly agnostic and independent of the Arduboy code, then a separate library should be created and maintained, totally independent and outside of the Arduboy library. In this case, _ab_ as a prefix to avoid namespace collisions would make even less sense. "_as_", for _Arduino Shapes_, or "_ap_", for _Arduino Physics_, would be more meaningful. Or use prefix _yyc514_, after the author. Maybe someday the library will be adopted as a built-in Arduino standard and the prefix will be changed or eliminated completely.

However, if the class/library is not Arduboy agnostic, it would be helpful if the namespace anti-collision prefix more clearly alluded to the Arduboy association.

joshgoebel:
> If the object creation class/library is truly agnostic and independent of the Arduboy code, then a separate library should be created and maintained, totally independent and outside of the Arduboy library. 

Right now my take is that @rogosher is entirely against this, unless I've misunderstood.  This also very clearly requires defining "what is Arduboy" and "what is Arduboy code"... I'd argue a lot of the current library has nothing to do with Arduboy at all.
- Drawing shapes isn't arduboy specific (and we have no specific ASM code that is)
- Frame management isn't arduboy specific, etc.
- Advanced button handling (which we don't have yet) isn't Arduboy specific

Etc.

This conversation happens elsewhere... I wish here we could get back on the top of how to introduce this huge breaking change to the codebase.  We can't really leave the old functionality there and depreciated or it kind of defeats the purpose of the new functionality - and prevents us from saving any space.

shogerr:
Haha, well my primary focus right now is just getting 1.0.0 published and accessible for the units going out. Sorry I haven't had time to construct hard opinions here. You are right dreamer, discussing naming semantics here is probably not the right place. I honestly don't have a full opinion or perspective on the PR itself (sorry). In about 5 days the bulk of the release work will be over... ha! Maybe we can move the library discussion to an issue?

MLXXXp:
> I wish here we could get back on the top of how to introduce this huge breaking change to the codebase.

I'd suggest we get the change fully defined, coded, tested and approved but not yet introduce it. Put it on a branch, waiting to be merged to the head.

Then, wait until whatever the concept of the _Arduboy Arcade_ turns out to be. Hopefully this happens more sooner than later.

Once the _Arcade_ is defined, created and working, we release the (non-backwards compatible) _printing separated_ version of the library, but only sanction games in the _Arcade_ that have been updated (if necessary) to work with the new library.

This will break any not converted games that use text, but the changes to fix them will be pretty straight forward and probably be doable easily and quickly by someone other than the original authors. Therefore, we should be able to get them updated and added to the _Arcade_ quite quickly. To help with this, we should try to get the word out on what has to be changed, so sketch authors or others can make updated versions available at a moments notice. We should provide a fairly detailed guide on what steps are required to accomplish the updates.

If possible, we should try to coordinate and synchronise things such that there are a good number of sketches converted, and the new library and _Arcade_ are released simultaneously.

The initial _Arcade_ may just be a web page with links to the sketches on GitHub (or some other place), but that's good enough to "sanction" games guaranteed to work with the new library.

This same concept could be used for other future library releases that "break everything". I.e. fix as many sketches as possible then release them simultaneously with the new library, and remove any remaining broken sketches (hopefully temporarily) from the _Arcade_.

joshgoebel:
Actually now that I think about it since half the games with their own copy of the libs aren't compiling anyways maybe it's a good time for some more breaking changes?

MLXXXp:
In addition to my last comment; let's say the old library is version 1.0 and the new one is version 2.0. We could fork the old library to its own repository and rename it to _Arduboy10_. Library version 2.0 would remain being called _Arduboy_

Anyone who really wanted to play a game removed from the _Arcade_, but didn't have the skills to update the game to work with the new library, could just install library _Arduboy10_ and change the `#include Arduboy.h` in the sketch to `#include Arduboy10.h`. They could then use this old sketch until a new updated version was sanctioned and available in the _Arcade_ again.

MLXXXp:
> Actually now that I think about it since half the games with their own copy of the libs aren't compiling anyways maybe it's a good time for some more breaking changes?

If we don't want to wait for an "official" arcade before introducing breaking changes, at least we should maintain a list, ideally with links, of games that work with the new library, so people don't waste time obtaining and uploading games that end up no longer working.

joshgoebel:
Starting to see the need for an official channel for games.  Having end users build from sources is TERRIBLE. :)

shogerr:
The 'arcade' solution would probably mirror choices made in this repo.

The arcade may be an in house solution, using [cloud9](https://c9.io/) gems, or simply a github index of projects... it's just too hard to say at this point. But we should probably keep in the back of our minds, that letting any arduboy user quickly and easily find, play and then edit or build a game is among the long term goals of the project. It's not this repositories responsibility to accomplish that, but this repo will hopefully enhance/expedite any solutions for application delivery built in the future.

joshgoebel:
Grrrr. SD card should have been a higher priority - and then you avoid all this messy source junk. :)

PR #84, opened 2016-02-13
flashlight mode

joshgoebel:
- hold up button when powering on

### DIFF
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -29,7 +29,9 @@ void Arduboy::begin()
 
   // flashlight
   if(pressed(UP_BUTTON)) {
-    sendLCDCommand(OLED_ALL_PIXELS_ON); // smaller than allPixelsOn()
+    // sendLCDCommand(OLED_ALL_PIXELS_ON); // smaller than allPixelsOn()
+    blank();
+    setRGBled(255,255,255);
     while(true) {}
   }
 
--- a/Arduboy.cpp
+++ b/Arduboy.cpp
@@ -27,6 +27,12 @@ void Arduboy::begin()
 {
   boot(); // required
 
+  // flashlight
+  if(pressed(UP_BUTTON)) {
+    sendLCDCommand(OLED_ALL_PIXELS_ON); // smaller than allPixelsOn()
+    while(true) {}
+  }
+
   // Audio
   tunes.initChannel(PIN_SPEAKER_1);
   tunes.initChannel(PIN_SPEAKER_2);
--- a/core.cpp
+++ b/core.cpp
@@ -250,9 +250,6 @@ void ArduboyCore::sendLCDCommand(uint8_t command)
   LCDDataMode();
 }
 
-#define OLED_PIXELS_INVERTED 0xA7 // All pixels inverted
-#define OLED_PIXELS_NORMAL 0xA6 // All pixels normal
-
 // invert the display or set to normal
 // when inverted, a pixel set to 0 will be on
 void ArduboyCore::invert(boolean inverse)
@@ -260,18 +257,12 @@ void ArduboyCore::invert(boolean inverse)
   sendLCDCommand(inverse ? OLED_PIXELS_INVERTED : OLED_PIXELS_NORMAL);
 }
 
-#define OLED_ALL_PIXELS_ON 0xA5 // all pixels on
-#define OLED_PIXELS_FROM_RAM 0xA4 // pixels mapped to display RAM contents
-
 // turn all display pixels on, ignoring buffer contents
 // or set to normal buffer display
 void ArduboyCore::allPixelsOn(boolean on)
 {
   sendLCDCommand(on ? OLED_ALL_PIXELS_ON : OLED_PIXELS_FROM_RAM);
 }
-
-#define OLED_VERTICAL_FLIPPED 0xC0 // reversed COM scan direction
-#define OLED_VERTICAL_NORMAL 0xC8 // normal COM scan direction
 
 // flip the display vertically or set to normal
 void ArduboyCore::flipVertical(boolean flipped)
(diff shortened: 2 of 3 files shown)


joshgoebel:
@rogosher Should put this in 1.0 release notes.

joshgoebel:
ROFL, maybe this should toggle the LED to white.

I will turn off lights and test.

MLXXXp:
Putting all three RGB LEDs full on, as well as setting all screen pixels on, would serve as a diagnostic for the LED as well as all the screen pixels. Perhaps you should also light both the TX and TX LEDs for the same purpose.

joshgoebel:
It's a flashlight, not a diagnostic. :)  And this is compiled code that always takes up room, it shouldn't try to do too much.

MLXXXp:
> It's a flashlight, not a diagnostic. :)

I'm just trying to come up with a better reason to justify it, to address your previous "Too gimmicky?" comment :wink:

"Holding down the _UP_ key during boot will light all the LEDS and turn on all pixels on the display, so you can verify if they're all working properly. It also makes a great flashlight!!"

MLXXXp:
> And this is compiled code that always takes up room

If you use the TXLED1 and RXLED1 macros, I think they only generate a few bytes. And you don't have to use PWM for the RGB LEDs, just a general purpose I/O output LOW, so they each should be similar to TX and RX, except you may have to also set the pins to outputs first.

joshgoebel:
> If you use the TXLED1 and RXLED1 macros, I think they only generate a few bytes. And you don't have to use PWM for the RGB LEDs, just a general purpose I/O output LOW, so they each should be similar to TX and RX, except you may have to also set the pins to outputs first.

Good points.  But I still don't want a hardware test.  That would be something else entirely.  Hardware test sounds like a sketch/demo kind of thing IMHO.

shogerr:
I like the diagnostic test idea. It's also a good point that it might be a nice source file as an example. Merging PR and adding to notes, maybe we need an issue crafted that deals with the creation of a diagnostic/test example.

MLXXXp:
I may be mistaken, but my testing shows that the _setRGBled()_ call generates over 400 bytes of code, if _setRGBled()_ isn't called within the sketch (which many sketches won't).

If so, as I said above, we should probably change this to using the pins as general purpose I/O, and setting them low using direct port writes, similar to what the TXLED0 and TX_RX_LED_INIT macros do.

joshgoebel:
> If so, as I said above, we should probably change this to using the pins as general purpose I/O, and setting them low using direct port writes, similar to what the TXLED0 and TX_RX_LED_INIT macros do.

File a sep issue, please (or even a PR).  The issue here is analogWrite is  heavy function, but you're right in this cause we don't need to PWM them we can just flip then on for the white.  What has to be done is the pins all set to output then setting them all LOW.

MLXXXp:
> File a sep issue, please (or even a PR).

Determining the proper way to control the RGB LED, via standard I/O at a low level,  looks like an interesting challenge. I'll try to do a PR if no one else does it, but it may not be for a few days. I don't think it's urgent, though. The code works as is. This would just be refactoring.

joshgoebel:
It's just find the port and bit position and set it to 0.  And you'll have to find the register that lets you set ports to output and make sure you mask it properly also.

Usually you can just read the digitalWrite source and follow it along to the right array to get the right ports, bits, masks, etc.

MLXXXp:
Yes, it's not difficult, but like you say, it needs a bit of investigation, plus testing and Git , etc.  I've got a few other things on my plate before I get to it, like re-wiring my home built system to be production version compatible including the RGB LED, plus non-Arduboy related stuff. I'll do it eventually if no one beats me to it.

I think it will likely turn out to be best to use the arrays in _pins_arduino.h_; _digital_pin_to_port_PGM[]_, _digital_pin_to_bit_mask_PGM[]_, etc., and using the _#define_s we have for the pin numbers; _RED_LED_, _GREEN_LED_, _BLUE_LED_ as indexes.

joshgoebel:
That's where to get the values but you can't use the arrAys. You need just the values. The arrays are part of the reason the compile size is so large. See the issue open on this topic. 

Ultimately we need defines for all out specific pin masks, bits, ports. Already started this with speaker stuff in my logo PR.

MLXXXp:
> That's where to get the values but you can't use the arrAys. You need just the values. The arrays are part of the reason the compile size is so large.

I was wondering about this. It's one of the things I want to take time to investigate. This is a problem when you use a **variable** as the index, but what if the index is a **constant**, as would be the case with using the pin number _#define_s?

The arrays are defined as _const uint8_t_ and the indexes would be constants, so there would be no variables in the expressions. Is the compiler smart enough to recognise this and boil it down to just a single value, instead of actually generating indexing code?

E.g. would these two statements generate identical code?

``` cpp
return digital_pin_to_bit_mask_PGM[RED_LED];

return 0x40;
```

MLXXXp:
OK, I just tried it. It looks like the compiler isn't that smart. Indexing the array with a constant generates 8 more bytes.

joshgoebel:
When you tell the compiler you want an array it gives you an array. Use that array somewhere and you're stuck with the whole thing.  I see where you were going with that logic though and I don't have any good specific reason that type of optimization couldn't be done (if the only detected reference was known at compile time), but yeah I don't think it's that smart yet.

Arduino has some really nice methods but unfortunately a lot of them are SUPER expensive because they aim to deal with ALL the pins/ports/timers on a chip even if we only care about one.

joshgoebel:
> The arrays are defined as const uint8_t and the indexes would be constants, so there would be no variables in the expressions. 

Const is still a variable, just a read-only variable.

joshgoebel:
The use of PROGMEM may be complicating this.

PR #85, opened 2016-02-13
buttonsState()

joshgoebel:


joshgoebel:
Rebased.

MLXXXp:
Calls to _getInput()_ within the library should now be changed to _buttonsState()_.

And is _buttonsState()_ clear enough that we can eliminate the _button_state_ variable in _pressed()_ and _notPressed()_? The new versions would be:

``` cpp
boolean Arduboy::pressed(uint8_t buttons)
{
  return (buttonsState() & buttons) == buttons;
}

boolean Arduboy::notPressed(uint8_t buttons)
{
  return (buttonsState() & buttons) == 0;
}
```

joshgoebel:
The intermediate variable could die now, yes.

Issue #70, opened 2016-02-09
FloatyBall or FlappyBall?


MLXXXp:
Just an observation:
The example sketch is named _FloatyBall_ but the intro screen says _FlappyBall_.

MLXXXp:
Also, we may want to get rid of the little _ARDUINO_ logo that's displayed on start up, especially if there's going to be an automatic _Arduboy_ logo.

MLXXXp:
Actually, I think this game should be removed unless it's tuned a bit. At the speed the ball rises and falls, I find it virtually unplayable.

shogerr:
It's not exactly a shining beacon of programming magnificence, but people do play it/ask for it.

I can fix those two issues, if we get the bootscreen included, the Arduboy logo can come out, thanks :+1:

joshgoebel:
I put it under proper frame management (at 30 FPS as it was designed for) but man you're right it's unplayable. :-)

joshgoebel:
To clean it up you'd make the bird a bitmapped sprite with animation frames and bump this to 60 FPS and then adjust the physics.  You could also have really nice bitmaps for the columns vs them being simple rects.  Lots of opportunity here once you fixed the physics.

shogerr:
This was discussed a bit with @bateske, and we can just take out flappy bird. It will probably be easily found on codebender etc in the future. Or an arduboy-extras repo.

PR #86, opened 2016-02-13
floaty ball example should use frame management

joshgoebel:


joshgoebel:
This also depends on buttonsState() being merged.

shogerr:
See issue #70, but I will merge this change and then we can address removing it for 1.0, and then putting the codebase in place where it can be reworked without being a priority for this repo.

joshgoebel:
Well does Arduboy itself want to host games that people can fork or do we want it gone gone and someone else responsibility completely?  I can understand why not every game would go into examples.

shogerr:
I would personally like to see it forked into another repository that Arduboy maintains in the end.

PR #87, opened 2016-02-14
Cleaned up HelloWorld and Buttons sketch examples

MLXXXp:
I know I shouldn't have spent time on this. I have more important things to do at the moment. But y'all seemed to want to still include these.

I cleaned up the code in PR #23, as per the discussion. I copied and pasted the source into a new branch, rather than squashing.
- I renamed helloworld.ino to HelloWorld.ino
- The sketches will only compile under IDE 1.5+ because I removed the _#include_s that only the library uses. This can easily be changed back if we decide to remain 1.0.x compatible. (I'm against it.)
- I changed _start()_ to _begin()_ and _clearDisplay()_ to _clear()_.
- I added boundary x, y position boundary checking in _Buttons.ino_
- In _Buttons.ino_, I reversed the _A_ and _B_ buttons. _B_ now moves the text up and _A_ moves it down, because _B_ is physically higher than _A_, so it's more intuitive.
- In _Buttons.ino_ I added _#define_s and math so that the length of the text string can be changed without any other changes needed to compensate.
- Lost of little cosmetic changes.

If we decide this is good to go, we should probably still wait for approval from @dragula96, since he's the original author.

### DIFF
--- a/examples/Buttons/Buttons.ino
+++ b/examples/Buttons/Buttons.ino
@@ -1,0 +1,106 @@
+/*
+Buttons example
+June 11, 2015
+Copyright (C) 2015 David Martinez
+All rights reserved.
+This code is the most basic barebones code for showing how to use buttons in
+Arduboy.
+
+This library is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
+License as published by the Free Software Foundation; either
+version 2.1 of the License, or (at your option) any later version.
+*/
+
+#include "Arduboy.h"
+
+// Make an instance of arduboy used for many functions
+Arduboy arduboy;
+
+// Variables for your game go here.
+char text[] = "Press Buttons!";
+byte x;
+byte y;
+
+// Width of each charcter including inter-character space
+#define CHAR_WIDTH 6
+
+// Height of each charater
+#define CHAR_HEIGHT 8
+
+// To get the number of characters, we subtract 1 from the length of
+// the array because there will be a NULL terminator at the end.
+#define NUM_CHARS (sizeof(text) - 1)
+
+// This is the highest value that x can be without the end of the text
+// going farther than the right side of the screen. We add one because
+// there will be a 1 pixel space at the end of the last character.
+// WIDTH and HEIGHT are defined in the Arduboy library.
+#define X_MAX (WIDTH - (NUM_CHARS * CHAR_WIDTH) + 1)
+
+// This is the highest value that y can be without the text going below
+// the bottom of the screen.
+#define Y_MAX (HEIGHT - CHAR_HEIGHT)
+
+
+// This function runs once in your game.
+// use it for anything that needs to be set only once in your game.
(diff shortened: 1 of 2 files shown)


joshgoebel:
I'm not sure how important it is to expose all the frame stuff for a static demo... honestly I'd probably just render the text once and then go into a while(true) idle() loop.  Frames serve no real purpose for a static page and really only get in the way of understanding the rest of the sketch, no?

That said I don't care that much, just wanted to throw my 0.02.  Take it or leave it. :)

MLXXXp:
> I'm not sure how important it is to expose all the frame stuff for a static demo... honestly I'd probably just render the text once and then go into a while(true) idle() loop.

If it were me, I'd just put all the code in _setup()_ and leave _loop()_ empty.

I got the feeling that, because this was an example for beginners, it was good to show that the frame stuff exists, and indicate that it's expected it would be used in more elaborate code. If a rookie wasn't aware of the technique, they may go on to just use delays and such.

MLXXXp:
I noticed I didn't add the _F_ trick for the string in _Buttons.ino_. Should it be done?

joshgoebel:
It's nice.  I just thought really you'd want a sequence of tutorials starting from simple and then getting more complex.  If you ask me the only thing that should be in the loop() is just frame management / clear /display and then there would be a separate gameloop() that loops calls.  Separate the system logic from the app logic.

examples/Buttons/Buttons.ino
```diff
@@ -0,0 +1,106 @@
...
+  
+  // set x and y to the middle of the screen
+  x = (WIDTH / 2) - (NUM_CHARS * CHAR_WIDTH / 2);
+  y = (HEIGHT / 2) - (CHAR_HEIGHT / 2);
+}
+
+
+// our main game loop, this runs once every cycle/frame.
+// this is where our game logic goes.
+void loop() {
```
shogerr, line 64: Could we stick to the style of the library here and do newlines before the curly-brace bracket?

Example:

```
void loop()
{
  ...
}
```

This would promote the same style throughout the repository, and it does make the function separation easier to read.

shogerr:
I like these examples, and think they demonstrate basic use. Not all programs/games will use the same design pattern, so these examples expose a user to the basics. 

See the line note, and then it can be merged.

examples/Buttons/Buttons.ino
MLXXXp, line 64: When you start the IDE, it creates a skeleton sketch with empty _setup()_ and _loop()_ functions.

These functions have the opening brace following the function name, on the same line. I changed the sketches to match, assuming the IDE template was the expected Arduino way to do it. Here the sketch I get when I start the IDE:

``` cpp
void setup() {
  // put your setup code here, to run once:

}

void loop() {
  // put your main code here, to run repeatedly:

}
```

shogerr:
Thanks, will merge :100:

PR #88, opened 2016-02-14
Change getInput() calls to buttonsState()

MLXXXp:
Also refactored pressed() and notPressed() to eliminate unnecessary local variable _button_state_.


joshgoebel:
Thanks.

PR #91, opened 2016-02-16
Move source files to src/ directory tree

MLXXXp:
For compliance with the[ Arduino IDE 1.5 Library Specification](https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification).

I made separate subdirectories each for _core_ and _audio_.

I didn't create an _extras_ directory. _README.md_, _LICENSE_, _CONTRIBUTORS.md_ and _TODO.md_ are all related directly to development, so I feel they should remain in the root.

We could add an _extras_ directory in the future, if desired, to include any user documentation beyond the _README.md_ file in the root.


joshgoebel:
Please don't go wild with this.  Not every single set of two files needs it's own subdirectory!

shogerr:
:+1:

MLXXXp:
> Not every single set of two files needs it's own subdirectory!

But what does it hurt? If you name the directory after the class, it makes it easy to remember the path for _#include_s.

I'm neutral on this, though. It's easy to move them all to a single directory. Just say the word and I'll do it. But I need to know, before I get more objections, is _utility_ an acceptable name for this single subdirectory, and if not, what name should it be?

joshgoebel:
I dunno.  I'd have left them in one directory until there was more of them and then decided how to best organize them.

shogerr:
I think the setup is OK currently. Core and Audio do enjoy their own folder because a user will be asking the Arduboy API to handle those apis, and they are submodules, or smaller api's to the larger set.

Keeping with the style guide hopefully helps clear this up. I do see how this could lead to many many sub directories when there should not be as more is added. But the readability of the project in it's current form is very well done, and is logical while falling within the style guide. Great contribution MLXXXP, much needed.

Let's open up a new issue if we want to discuss what to do going forward with the file organization.

MLXXXp:
> I do see how this could lead to many many sub directories when there should not be as more is added.

Just how many more sub-API's do you think will be added? I think that many extensions would turn out to be better handled by separate, independent libraries. These new libraries would just  use the API functions provided by Arduboy.h, much like the Arduboy library _#include_s and uses the EEPROM and SPI libraries.

shogerr:
Sold!

PR #92, opened 2016-02-17
separate PR for playing with boot noise

joshgoebel:


shogerr:
Attempt on devkit:
Boot logo scrools, at the end of logo a faint click is heard.
Expected? A tone.

I then loaded up the tunes example, heard the same click at the end of the logo loading sequence, then let the composition stored play through. It played Bach's Symphony no. 12, So everything seems to work fine as far as playing audio.

This seems relevant. 
https://github.com/yyyc514/ArduboyLib/blob/hw_boot_sound/src/Arduboy.cpp#L82

src/Arduboy.cpp
```diff
@@ -26,7 +27,32 @@ void Arduboy::start() // deprecated
 void Arduboy::begin()
 {
   boot(); // required
+  bootUtils();
 
+  bootLogo();
+
+  // Audio
+  tunes.initChannel(PIN_SPEAKER_1);
```
shogerr, line 35: Wrong pins for devkit?

joshgoebel, line 35: The dev kit pins didn't change in the code.  Are you compiling with DEVKIT and not PROD?

shogerr, line 35: Yeah, set

``` c++
//#define ARDUBOY_10   //< compile for the production Arduboy v1.0
#define AB_DEVKIT    //< compile for the official dev kit
#endif
```

It may be the tone doesn't generate the same sound when ran through the different unit's sound hardware.

MLXXXp:
> This seems relevant.
> https://github.com/yyyc514/ArduboyLib/blob/hw_boot_sound/src/Arduboy.cpp#L82

That could be true. The DevKit shorts two pins together and connects them to one speaker lead. The other speaker lead goes to ground. By setting both speaker pins as outputs on a DevKit, you're shorting two outputs together (almost always a bad thing). Pin 2 remaining low may be holding Pin 1 low when it tries go high.

As an experiment, try commenting out the following line in function _Arduboy::bootLogo()_:

``` cpp
pinMode(PIN_SPEAKER_2, OUTPUT);
```

On the production Arduboy, the speaker leads are connected to one pin each, so you have to set them both to outputs to get any sound when you toggle one of them.

joshgoebel:
> By setting both speaker pins as outputs on a DevKit, you're shorting two outputs together (almost always a bad thing). Pin 2 remaining low may be holding Pin 1 low when it tries go high.

I'd have thought the same thing exactly the tunes code sets both pins to OUTPUT when it inits and that seems to work just fine... so... I'm lost.

MLXXXp:
@yyyc514, do you get this same problem on your DevKit as @rogosher is reporting with his?

I don't have time tonight, but I can try it tomorrow with my breadboarded DevKit equivalent system, which allows me to separate the two speaker pins.

joshgoebel:
Dunno, haven't plugged it in.  I don't play with them every night. :-)  Been working on other things.  So could be a legit problem overall, dunno.

rogosher:
Double check that `ab_logo.h` is in the right place, and then we can merge this.

joshgoebel:
Wow, we hear totally diff things.

> Kevin: its fine for now but i wanna try and make something cooler

I don't want to be the person who chose the placeholder sound.  And it sounds like we still have issues on Devkits anyways.  Closing this.  Kevin says we'll try and figure out an official sound in the next week or two.

PR #96, opened 2016-02-18
Change README.md

MLXXXp:
Fix the _libraries_ directory location for Linux.
Add how to use the IDE menus to include the library in a sketch.

**Note**:  The _README.md_ file has DOS CR/LF line termintors. Do we want to change it to Unix LF terminators, to be consistent with the rest of the library? If so, I can add another commit with it changed, before the PR is merged.


shogerr:
The `Documents` folder is where it goes in linux. It works for myself at least and is in the documentation.

https://www.arduino.cc/en/Guide/Libraries#toc5

I'd imagine it picks up both library locations, to handle those users who never install X. However, as `Documents` is created in a user's home with most X installs, and it's what the documentation calls for, we should probably leave it? Maybe we can add a line about non X installs for completeness? Be nice to find it in documentation that it supports both though.

The line about adding the define through the gui is nice, a good addition.

Use Unix line endings here as well.

MLXXXp:
Since the Ubuntu Linux _Software Center_ only provides IDE version 1.0.5, I installed 1.6.7 on my Ubuntu 15.04 system using [this guide](http://ubuntuhandbook.org/index.php/2015/11/install-arduino-ide-1-6-6-ubuntu/). The result was that it created the _Arduino_ folder directly under my _Home_ folder, not in _Documents_. So, it looks like the default isn't standard for all Linux distributions and installations.

> I'd imagine it picks up both library locations, to handle those users who never install X.

No, the IDE uses whatever is specified in the _Sketchbook location_ field, found in the _Settings_ tab from menu _File -> Preferences_. A user can change it to whatever they wish.

How about if I change _README.md_ to:

---

**Linux**

```
/home/username/Documents/Arduino/libraries
```

or

```
/home/username/Arduino/libraries
```

---

and at the end of the entire list I'll add:

---

If you don't find the `libraries` folder in one of the above locations, you can determine its location using the navigation menu `File > Preferences`. In the `Settings` tab will be a `Sketchbook location:` field. The `libraries` folder will be in the folder set in this field.

---

shogerr:
Ok, the addition at the bottom works pretty well. If a note about where to find the `libraries` location is included it makes sense to drop the `/home/username/Documents/Arduino/libraries` bit and follow what the IDE defaults to and simply include `/home/username/Arduino/libraries` for Linux. Shorter is probably better in the case of this readme.

MLXXXp:
> If a note about where to find the libraries location is included it makes sense to drop the /home/username/Documents/Arduino/libraries bit and follow what the IDE defaults to and simply include /home/username/Arduino/libraries for Linux.

But if your Linux system has it under _Documents_ and [_Installing Additional Arduino Libraries_](https://www.arduino.cc/en/Guide/Libraries#toc5) says that's where is should be, I'm wondering if that location is more common. I think maybe we should use it and delete where I have it, if you only want to give one.

The additional note will help people find it, anyway.

shogerr:
<strike>Right on, I was reading through the changelogs quickly to see if there was anything mentioned. Nothing so far. I can boot up a VM and test it in Debian and Ubuntu, from the package manager and other ways, assuming there is a package for the IDE. I'll check the default locations.
</strike><strike>
Of course every distribution is prone to do their own thing when it comes to file locations and third party applications. It might just be worth targeting Ubuntu, Debian, Centos and Fedora for tests.
</strike><strike>
Let's hold off a second, and I'll do a couple vagrant boxes, at least one for Ubuntu, and see where the  libraries default to installing from the package manager, and then I'll try manually installing the IDE and checking.</strike>

Let's just use the `Documents/` folder. :smile_cat:

Issue #94, opened 2016-02-18
ab_logo.h shouldn't  be included in sketches


MLXXXp:
When a developer uses _Sketch -> Include Library -> Arduboy_ from the IDE menus, `#include <ab_logo.h>` is added to the sketch (along with `#include <Arduboy.h>`). This header shouldn't be included in a sketch.

The reason _ab_logo_ is included is because the IDE includes all _.h_ files residing directly in the _src_ directory.
Possible ways to fix this:
- Move _ab_logo.h_ to a subdirectory under _src/_ (such as _bitmaps_) and in _Arduboy.cpp_ change `#include "ab_logo.h"` to `#include "bitmaps/ab_logo.h"`
- Change _ab_logo.h_ to _ab_logo.c_, as _glcdfont.c_ is. Why is one a _.h_ file and one a _.c_ file anyway? They both do the same thing, so should have the same extension. The library should be consistent.

Note that if we decide that _glcdfont.c_ should be _glcdfont.h_, it should be moved to a subdirectory under _src/_ as well, for the same reason. It can be the same directory that we put _ab_logo.h_ in.

shogerr:
I do like moving from a header tile to a source file, as I don't see this file as being something many people will change the source for, and it's a small bit of source.

MLXXXp:
So should _glcdfont.c_ be changed to _glcdfont.h_?

Is _bitmaps_ a good name for the directory to put both files in?

**EDIT**: Sorry, I read "I do not like" when you wrote "I do like". Disregard this comment.

joshgoebel:
One difference is C files are going to be auto compiled and .h files are not, but since both of these files only add a variable that is either referenced or unreferenced I don't think it should really matter to the compiler... so I think them both being .c wouldn't necessarily be harmful... but I really don't know the "proper" C opinion on this.  It seems in most of the source I find/read that large amounts of data/images/etc are put in the headers.

I think we should find out and then either rename both, or make both .h.

I preset `assets` or `resources`.  Bitmaps is too specific.

joshgoebel:
If we're still getting carried away with dirs maybe put the font in `fonts`. :-)  I'd love to see us ship a non ugly font, but no time to work on it.

yyyc514:
I'd love to see us ship a non ugly font, but no time to work on it.

MLXXXp:
[...]

PR #98, opened 2016-02-19
Rename ab_logo.h to ab_logo.c

MLXXXp:
After rename, modified to work as a .c file

As discussed in issue #94

### DIFF
--- a/src/Arduboy.cpp
+++ b/src/Arduboy.cpp
@@ -1,6 +1,6 @@
 #include "Arduboy.h"
 #include "glcdfont.c"
-#include "ab_logo.h"
+#include "ab_logo.c"
 
 Arduboy::Arduboy()
 {
--- a/src/ab_logo.c
+++ b/src/ab_logo.c
@@ -1,5 +1,7 @@
-#ifndef AB_LOGO_H
-#define AB_LOGO_H
+#include <avr/pgmspace.h>
+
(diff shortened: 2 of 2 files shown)


MLXXXp:
@yyyc514, that's funny... The same two thoughts popped into my head as I was waking up this morning :smile: 

> Are the ifdefs really needed at all now?

No, but they aren't hurting anything either. I can add a commit to this PR to remove them if we want. The same goes for the ones in _glcdfont.c_. Also, `#include <avr/io.h>` isn't required in _glcdfont.c_. Do we want to clean up all of these?

> And I don't think Arduboy needs to include the .c file... isn't it going to get compiled by default?

From experimenting, I've found that including _ab_logo.c_ (and _glcdfont.c_) **is** required in _Arduboy.cpp_. Otherwise you get an error like:
`Arduboy.cpp:72:22: error: 'arduboy_logo' was not declared in this scope`

There's evidence that _ab_logo.c_ is compiled but I suspect that the object file doesn't get linked in if not _#include_d. Or maybe it's linked, but in _Arduboy.cpp_ it can't be determined what _arduboy_logo_ is.

Another way to make it work is by adding a declaration in _Arduboy.cpp_:

``` cpp
extern PROGMEM const unsigned char arduboy_logo[];
```

but I think just the `#include "ab_logo.c"` works well enough.

PR #101, opened 2016-02-20
`boot()` is the minimal boot mode to save space

joshgoebel:


MLXXXp:
I think _begin()_ should be changed to call _beginMinimal()_ and then call all of the available feature functions. That way, a sketch can exactly emulate _begin()_ by using _beginMinimal()_ and then calling the same functions that _begin()_ does, in the same order. If it's OK for a sketch to call all the feature functions after _beginMinimal()_ then they can go at the very end of _begin()_ as well.

As it is, the sketch doesn't get the same flashlight capability as with using _begin()_. The sketch would have to add code to check for the _UP_ button.  The _flashlight()_ function should be changed back to the way it was before, which checked for _UP_. I think its OK to say that if a sketch wants an automatic-on flashlight later on in its code, it can just control the RGB LED itself. We could just document how to do a flashlight in a tutorial somewhere.

My reasoning is that a sketch developer will say:
"I'm running low on space so I'm going to use _beginMinimal()_, but just getting rid of the logo is enough. I can keep the flashlight and the system control menu."
So the sketch would contain:

``` cpp
  arduboy.beginMinimal();
  arduboy.flashlight();
  arduboy.sysControl();
```

joshgoebel:
The button control sequence being outside the function is intentional.

If you factor out the audio (or decide the order truly doesn't matter - after some thought it should not) then someone who wants to do this already had a minimal boot, they just need to call the hardware boot() and then build whatever they want on top of that.  There isn't need to build more functionality here.

joshgoebel:
`audio.begin()` is still a tiny wrinkle, but not a problem.  Most sketches I see just call on() or off() anyways (since we have no great way yet to configure audio) - and on() now does a lazy startup if necessary.

I think the audio part will become clearer when we decide what to do about tunes and synth long-term.

MLXXXp:
OK, I see your point of view and don't have a problem with this PR except:

Do you think we need to deprecate _bootNoLogo()_? I doubt anyone is using it yet or even knows it exists. I say remove it now and then try to help out if someone complains that it disappeared. I think it would be easier than trying to decide when it's safe to fully remove it in the future, and I don't think deprecated functions should be left in forever. The reason _start()_ and _clearDisplay()_ were deprecated is because they were already in heavy use, so it may have taken a while to fix all the broken sketches (I'd like to see them removed as soon as possible, as well).

Add a note in the _begin()_ Doxygen comments in _Arduboy.h_ like:

> If necessary to gain some code space, you can use boot() instead of begin() for a minimal system. See the boot() documentation for more info.

Then, put Doxygen documentation similar to what you've written for _bootNoLogo()_ under _boot()_ itself.

Another thing:
In order for a sketch to be able to use _boot()_ as a minimal base, anything that _begin()_ does will have to be available to the sketch. Future additions to _begin()_ can't use private functions or variables or anything else that the sketch wouldn't  be able to do.

In other words, it should be possible for a sketch to copy anything in _begin()_ and have it compile and execute exactly the same. (Of course, object instance names would have to be added, e.g. `audio.begin()` changed to `arduboy.audio.begin()`, but that should be about it.)

We should add a note to the comments at the start of _begin()_, in _Arduboy.cpp_ (not to the Doxygen comments in the header file), spelling out this requirement, for the benefit of future developers working on _begin()_.

joshgoebel:
>  bootNoLogo()

We could remove it.

> Future additions to begin() can't use private functions or variables or anything else that the sketch wouldn't be able to do.

I get that.  If something appears truly private then we'd change our take here and encourage people to sub-class I guess.

src/Arduboy.h
```diff
@@ -41,17 +41,35 @@ class Arduboy : public Print, public ArduboyCore
...
+   * need and start there.  Calling just `boot()` might work also
+   * depending on your requirements.
+   *
+   * The minimum recommended `begin` replacement:
+   *
+   *   arduboy.boot()         // raw hardware init
+   *   arduboy.audio.begin()  // if you need audio
+   */
+  // void boot(); // defined in core.cpp
+
```
MLXXXp, line 63: Will putting Doxygen comments on a commented out function generate proper documentation when you use Doxygen to create it?

joshgoebel, line 63: No idea, I don't use Doxygen.

MLXXXp, line 63: Then I guess the next question is "Do we, or do we intend to, use [Doxygen](http://www.stack.nl/~dimitri/doxygen/) to automatically generate documentation for the libaray?". The comments for functions in the header file have been formatted for it.

joshgoebel, line 63: I don't care.  That was @davidperrenoud thing.  To me it's useless if we aren't generating HTML and publishing it.  I believe in the importance of writing good comments so I try to do that regardless.

MLXXXp, line 63: You don't care but does anyone else, or has the idea been totally abandoned?

> To me it's useless if we aren't generating HTML and publishing it.

Is it not a good idea to publish library documentation somewhere on _arduboy.com_? I found that Doxygen did a good job for the [MicroView Library](http://learn.microview.io/doco/html/annotated.html), which is similar.

joshgoebel, line 63: No one is going to care if we don't publish it - I'm not sure it serves any point at all if we don't publish it - but that's not my realm - that's @davidperrenoud I think.  Published library docs would of course be a nice thing.

MLXXXp:
I just noticed that _boot()_ is a protected function. It has to be public if it's going to be called in a sketch.

MLXXXp:
While you're in clean up mode, how about renaming the remaining things containing "LCD", like you did by changing _bootLCD()_ to _bootOLED()_

In core: `LCDDataMode()`, `LCDCommandMode()`, `sendLCDCommand()`

joshgoebel:
> In core: LCDDataMode(), LCDCommandMode(), sendLCDCommand()

Didn't want to have to deprecate them all - they are public.  The ones I touched are internal.  And I can't decide long term if it should be OLED or SSD1306 or something.  We aren't really talking to the OLED, we're talking to the controller.  So until I 100% sure of a name for public methods I thought I'd leave them alone.

> Do we need this compile for 8MHz capability? 

I think it's nice and I'm the one who added it.  It certainly doesn't hurt anything now that I already did the work of getting it there.  IMHO it should be added to the board file and available for any game that uses <50% of CPU.  It's a great battery savings.

MLXXXp:
> IMHO it should be added to the board file and available for any game that uses <50% of CPU. It's a great battery savings.

Wouldn't it be better to have a function call? The sketch could slow itself down and then stay that way, or speed back up if required at some point.

joshgoebel:
> Wouldn't it be better to have a function call? The sketch could slow itself down and then stay that way, or speed back up if required at some point.

Less flash storage to do it at compile time - PLUS changing speeds during runtime break all sorts of timers and other timing specific things - the Arduino software really doesn't like you doing that.  It wants to know at compile time what peed it's going to be running and a lot of things are set using compile time defines.

src/Arduboy.h
davidperrenoud, line 63: I agree with @yyyc514, keeping the comments is a good idea. And I indeed formatted the comments for Doxygen but @rogosher has yet to publish it.

MLXXXp:
Can I urge you then, to not use the _F_CPU_ value directly to decide to slow down the CPU clock? Use a new define, such as `#define ARDUBOY_SET_F_CPU 8000000L`, which is made available to the library when it's compiled, the same way that ARDUBOY_10 or AB_DEVKIT will be. The value of F_CPU should only be used for calculations or settings that are dependent on the clock speed, not, on its own, to actually change the speed.

The reason is that someone may want to use the Arduboy library with a compatible system that actually physically runs at 8MHz, such as one based on a SparkFun Pro Micro 3.3V/8MHz (like I once did). The _boards.txt_ file for the Pro Micro would contain `promicro8.build.f_cpu=8000000L` so if the library just tests _F_CPU_ it would slow the clock down to 4MHz. This issue can be alleviated by using a separate _#define_, that is only set by choosing a board type that indicates you want a speed change via software.

The library would do something like:

``` cpp
#if defined ARDUBOY_SET_F_CPU && ARDUBOY_SET_F_CPU != F_CPU
  #if ARDUBOY_SET_F_CPU == 8000000L
  bootCPUSpeed();
  #endif
#endif
```

Also, I'd rename _bootCPUSpeed()_ to something like _bootCPUSpeed8MHz()_  or have it accept an argument, in case we want to be able to select other lower speeds, such as 4MHz, for puzzle games or other nearly static sketches.

MLXXXp:
@davidperrenoud, you wrote:

> I agree with @yyyc514, keeping the comments is a good idea.

I wasn't suggesting that the comments be removed. I was questioning if Doxygen could, and if so, how it would, deal with a Doxygen formatted comment block in front of the commented out first line of a function that exists in a different file. Specifically:

_Arduboy.h_ contains:

``` cpp
  /// Init just hardware, no logo, no boot utilities.
  /**
   * Look at the source for `begin()` and just rip out what you do not
   * need and start there.  Calling just `boot()` might work also
   * depending on your requirements.
   *
   * The minimum recommended `begin` replacement:
   *
   *   arduboy.boot()         // raw hardware init
   *   arduboy.audio.begin()  // if you need audio
   */
  // void boot(); // defined in core.cpp
```

As indicated, _boot()_ itself is defined in file _core.h_ with it's own different Doxygen formatted comments. So, can you tell me what Doxygen is going to do with the above lines? They are intended to be read by someone using the library. Will they be included as part of what Doxygen generates?

joshgoebel:
> Can I urge you then, to not use the F_CPU value directly to decide to slow down the CPU clock?

Please raise this as a sep issue.

PR #106, opened 2016-02-21
rip out tunes

joshgoebel:


joshgoebel:
Do we want to move the tunes example also?  I'm actually OK with leaving it here.

PR #107, opened 2016-02-23
faster drawFastHLine from Adafruit SSD1306 lib

joshgoebel:


src/Arduboy.cpp
```diff
@@ -385,10 +385,47 @@ void Arduboy::drawFastVLine
...
+    w = (WIDTH - x);
+  }
+
+  // if our width is now negative, punt
+  if (w <= 0) {
+    return;
+  }
+
+  // buffer pointer plus row offset + x offset
+  register uint8_t *pBuf = sBuffer + ((y/8) * WIDTH) + x;;
```
shogerr: I haven't tried compiling this yet, looks like a typo. Or maybe I need some education?

MLXXXp: If you mean the double semicolon, it's probably a typo. The second semicolon will just be a null statement and do nothing. It will compile correctly and produce exactly the same code as if there was only one semicolon.

It should probably be fixed, though.

joshgoebel: Yeah, ;; is not a problem.  Just an empty statement.

Issue #108, opened 2016-02-23
Speaker needs both pins set to OUTPUT


atomic-salamander:
Both speaker pins must be set to OUTPUT, or neither will work, like so:

```
pinMode(PIN_SPEAKER_1, OUTPUT);
pinMode(PIN_SPEAKER_2, OUTPUT);
```

This is already done in the Arduboy audio library but if someone wants to make their own and doesn't plan on using both pins it can be very difficult to track down why the speaker isn't working.

MLXXXp:
This is only true for the production Arduboy. The DevKit has the two pins tied to one speaker terminal, with the other terminal tied to ground. If you set both pins to outputs on the DevKit, the processor could be damaged.

To address the problem of someone using a different audio library, it would probably be a good idea to initialize the speaker pins in _boot()_. The production Arduboy should have the pins set to ouputs, and the DevKit to inputs (they should default to inputs, but it can't hurt to set them again, to make sure). I've submitted PR #109 to do this.

The _ArduboyAudio_ class should probably be modified to only allow one channel for the Devkit.

joshgoebel:
@atomic-salamander You're sketch should still be calling Arduboy.audio.begin() so that the EEPROM registered audio/volume settings can be honored.  If you call audio.begin() then the pins should be setup for you. (if the audio is enabled)  

If you have your own toggle code you want audio.on() and audio.off().

joshgoebel:
This will be a bit more obvious after https://github.com/Arduboy/Arduboy/pull/106 but audio is already separate from tunes. (other than currently sharing the same files)

joshgoebel:
Are there any power implications to have the pins OUTPUT vs INPUT?

MLXXXp:
Setting the speaker pins both to outputs (on the production Arduboy) could possibly draw less power than with either of  them being an input. The reason is that the piezo element can work as a microphone as well as a speaker. If the "speaker" picks up noise it could generate AC voltages on the pins. If this voltage falls somewhere between a high and a low on an input, this input could draw current internally.

Even if the speaker doesn't generate an AC voltage, it appears as a DC high impedance, so the pins are essentially unconnected. The ATmega32U4 datasheet has this to say about unconnected inputs:

> **10.2.6 Unconnected Pins**
[...]

PR #109, opened 2016-02-23
Init speaker pins in boot()

MLXXXp:
In case someone uses _boot()_ to use their own audio system, and so doesn't call _audio.begin()_.

As discussed in issue #108.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -13,6 +13,15 @@ const uint8_t PROGMEM pinBootProgram[] = {
   PIN_A_BUTTON, INPUT_PULLUP,
   PIN_B_BUTTON, INPUT_PULLUP,
 
+  // speaker
+#ifdef ARDUBOY_10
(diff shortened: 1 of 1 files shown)


shogerr:
This looks ok to me. Merge if satisfied @yyyc514. I will merge tomorrow if it's still open. : )

joshgoebel:
This does not belong in boot.  Currently Arduboy.audio.begin/on() handles this and 3rd party sfx libraries should still use ArduboyAudio.  That's is what allows us to have a boot tool (that you are working on) to turn sound on/off, etc, globally.  Doing a custom boot you still need to respect the audio settings - and if you don't want to respect them then you need to setup these pins yourself.

If they use audio properly then they get this init for free, if they don't then they are on their own - and they are outside the guidelines anyways.  I've been thinking of a way to handle system mute across multiple tunes library options and this is great.  Just turn off the audio pins if audio.mute()/off() is called.  That's kind of a rough trick, but it would get the job done.

PR #111, opened 2016-02-24
Use a dedicated #define to adjust the CPU clock

MLXXXp:
As discussed [here](https://github.com/Arduboy/Arduboy/pull/101#issuecomment-187314617).

I moved the speed adjustment code from _boot()_ to the constructor. The sooner it gets done, the better. Ideally, it should be done immediately after the bootloader exits.

I added the capability of setting 4MHz, just because. It doesn't hurt anything.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -89,14 +89,46 @@ const uint8_t PROGMEM lcdBootProgram[] = {
 };
 
 
-ArduboyCore::ArduboyCore() {}
+ArduboyCore::ArduboyCore()
+{
+// ARDUBOY_SET_F_CPU wil be set by selecting a lower speed in the IDE
+#if defined ARDUBOY_SET_F_CPU && ARDUBOY_SET_F_CPU != F_CPU
+  #if ARDUBOY_SET_F_CPU == 8000000L
+    setCPUSpeed8MHz();
+  #elif ARDUBOY_SET_F_CPU == 4000000L
+    setCPUSpeed4MHz();
+  #endif
+#endif
+}
+
+#if defined ARDUBOY_SET_F_CPU && ARDUBOY_SET_F_CPU != F_CPU
+  #if ARDUBOY_SET_F_CPU == 8000000L
+// if we're compiling for 8MHz we need to slow the CPU down because the
+// hardware clock on the Arduboy is 16MHz
+void ArduboyCore::setCPUSpeed8MHz()
+{
+  uint8_t oldSREG = SREG;
+  cli();                // suspend interrupts
+  CLKPR = _BV(CLKPCE);  // allow reprogramming clock
+  CLKPR = 1;            // set clock divisor to 2 (0b0001)
+  SREG = oldSREG;       // restore interrupts
+}
+  #elif ARDUBOY_SET_F_CPU == 4000000L
+// if we're compiling for 4MHz we need to slow the CPU down because the
+// hardware clock on the Arduboy is 16MHz
+void ArduboyCore::setCPUSpeed4MHz()
+{
+  uint8_t oldSREG = SREG;
+  cli();                // suspend interrupts
+  CLKPR = _BV(CLKPCE);  // allow reprogramming clock
+  CLKPR = 2;            // set clock divisor to 4 (0b0010)
+  SREG = oldSREG;       // restore interrupts
+}
+  #endif
+#endif
 
 void ArduboyCore::boot()
 {
-  #if F_CPU == 8000000L
-  bootCPUSpeed();
-  #endif
-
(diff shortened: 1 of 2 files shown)


joshgoebel:
This is broken.  These tests needs to be based on the REAL F_CPU value, and that needs to change.  There are timing related assembly scattered all throughout the Arduino libs that depends on knowing the CPU value the CPU will be running at.  So any board file that runs at 8Mhz (or 4Mhz) would need to pass in the correct F_CPU value and be compiled for that speed or a lot of routines would not work properly.

So what the define should look like is:

```
#if F_CPU == 8000000L && defined(ARDUBOY_HARDWARE)
```

Then the board file needs to set a specific flag to indicate that Arduboy board is being used vs say Leonardo. I suggest ARDUBOY_HARDWARE, but open to suggestions.  Perhaps one is already set based on the device name?  I'd have to test compiling with the boards file again.

I don't think anyone needs 4Mhz, so lets drop that... it's too slow for most games and doesn't have many real advantages over sleeping at 8Mhz.  I only added 8Mhz so those who were paranoid could run the hardware within the rated hardware specs (which specify 8Mhz for the low voltage we are using).

This wouldn't be accepted until ARDUBOY_HARDWARE was also added to the boards file (or if there is already another define we can use that the custom boards file is setting).

joshgoebel:
Your prior point about dropping 8Mhz is probably valid, but I'm attached to it and I think it could serve a purpose for some sketches, so since it's already in there I vote it stays. That doesn't mean we need to go and add all the other possible slow clock speeds that COULD be supported.

We support overclocked 16 and with-in spec 8.  I don't see the need to add others unless a specific need for them becomes apparent.

MLXXXp:
> This is broken.

Yes, you're right. My mistake :flushed: .  As you say, we don't want a new _#define_ to indicate the desired clock speed; we want to know that the hardware is actually physically running at 16MHz, regardless of what F_CPU is set for, so we can compensate.

> I suggest ARDUBOY_HARDWARE, but open to suggestions. Perhaps one is already set based on the device name?

Well, there will be `ARDUBOY_10` or `AB_DEVKIT` but we'd have to test for both, and possibly update for any new variants added in the future. I'd just use a _#define_ dedicated to the purpose. Instead of _ARDUBOY_HARDWARE_, I suggest we make it a bit clearer and future proof. I'd call it _ARDUBOY_HARDWARE_16MHZ_ or if too long, _ARDUBOY_16MHZ_.

> I don't think anyone needs 4Mhz, so lets drop that.

That's fine with me. Like I said, it doesn't hurt anything, but if you feel it's of no value, it can go.

Let me know if _ARDUBOY_HARDWARE_16MHZ_ is OK or if you want something else. If everything else is fine, including other names I've used, and doing the work in the constructor, then I'll fix this PR or make a new one.

**Note**: In [_boards.txt_](https://github.com/yyyc514/hardware_arduboy/blob/master/avr/boards.txt) you should change all _Mhz_ used as text or info, to _MHz_. The _H_ should be capitalized.

joshgoebel:
> ARDUBOY_HARDWARE_16MHZ or if too long, ARDUBOY_16MHZ.

I had the same though, that that would be clearer. But what about being explicit.  `ARDUBOY_16MHZ_HW_CLOCK`.

> ARDUBOY_10 or AB_DEVKIT 

These can't be used as  ARDUBOY_10 will be defaulted even if nothing else is passed, so it would still be defined on a Leonardo. 

> and doing the work in the constructor

I don't think it matters that much, I prefer init functions vs the constructor (gives the user more power).  Someone would have to make a point of doing a lot of timing specific work first before calling `begin()` for this to even matter.  Not something I've ever seen in a sketch.  I'm ambivalent.

joshgoebel:
> Note: In boards.txt you should change all Mhz used as text or info, to MHz. The H should be capitalized.

No objection if they aren't required to be that way for some funny reason, probably just typos.  Make a PR.

MLXXXp:
> > and doing the work in the constructor
> 
> I don't think it matters that much, I prefer init functions vs the constructor (gives the user more power).

But I think in this case we don't want to empower the user. As you've said, it must be done and then not touched after that, if we want all the internal timing to work properly. And as I said, it would be even better if it could be done before the sketch even enters _setup()_, which would give the user no control of it whatsoever.

I'm wondering if it could be done before sketch execution by compiling and linking in some code using the capabilities of _platform.txt_ and/or _boards.txt_, with the source code residing somewhere under _hardware_ along the same path as those _.txt_ files. This would make the speed change work even if the Arduboy library wasn't used. I haven't the time to research this right now.

MLXXXp:
Closed. Replaced by PR #112.

PR #112, opened 2016-02-25
Use a dedicated #define to adjust the CPU clock (fixed)

MLXXXp:
As per discussion it PR #111.

I've tested this as best as possible without a proper boards.txt (which I haven't had time to install and get working).

Regarding the [boards.txt](https://github.com/yyyc514/hardware_arduboy/blob/master/avr/boards.txt) file that @yyyc514 designed to allow selecting this:
- As mention previously, all instances of _Mhz_, that will displayed to the user or are used as comments in the file, should be changed to _MHz_. The _H_ should be capitalized.
- I think using `Speed`, as a menu item to select the operating speed, is a bit ambiguous and could be confusing. Users could think it specifies the physical clock speed and assume that there must be an 8MHz version of the Arduboy out there, which you would select 8MHz for. I would suggest changing the menu item `Speed` to `Run Arduboy` and change the selection values to `at full speed, 16MHz` and `with speed reduced to 8MHz`.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -89,14 +89,29 @@ const uint8_t PROGMEM lcdBootProgram[] = {
 };
 
 
-ArduboyCore::ArduboyCore() {}
+ArduboyCore::ArduboyCore()
+{
+// ARDUBOY_16MHZ_HW_CLOCK will be set by the IDE using boards.txt
+#if F_CPU == 8000000L && defined(ARDUBOY_16MHZ_HW_CLOCK)
+  setCPUSpeed8MHz();
+#endif
+}
+
+#if F_CPU == 8000000L && defined(ARDUBOY_16MHZ_HW_CLOCK)
+// if we're compiling for 8MHz we need to slow the CPU down because the
+// hardware clock on the Arduboy is 16MHz
+void ArduboyCore::setCPUSpeed8MHz()
+{
+  uint8_t oldSREG = SREG;
+  cli();                // suspend interrupts
+  CLKPR = _BV(CLKPCE);  // allow reprogramming clock
+  CLKPR = 1;            // set clock divisor to 2 (0b0001)
+  SREG = oldSREG;       // restore interrupts
+}
+#endif
 
 void ArduboyCore::boot()
 {
-  #if F_CPU == 8000000L
-  bootCPUSpeed();
-  #endif
-
   SPI.begin();
   bootPins();
   bootOLED();
@@ -108,19 +123,6 @@ void ArduboyCore::boot()
 
   bootPowerSaving();
 }
-
-#if F_CPU == 8000000L
-// if we're compiling for 8Mhz we need to slow the CPU down because the
-// hardware clock on the Arduboy is 16MHz
-void ArduboyCore::bootCPUSpeed()
-{
-  uint8_t oldSREG = SREG;
-  cli();                // suspend interrupts
-  CLKPR = _BV(CLKPCE);  // allow reprogramming clock
-  CLKPR = 1;            // set clock divisor to 2 (0b0001)
-  SREG = oldSREG;       // restore interrupts
-}
-#endif
 
 void ArduboyCore::bootPins()
 {
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -284,9 +284,9 @@ protected:
     void static inline safeMode() __attribute__((always_inline));
 
     // internals
+    void static inline setCPUSpeed8MHz() __attribute__((always_inline));
     void static inline bootOLED() __attribute__((always_inline));
     void static inline bootPins() __attribute__((always_inline));
-    void static inline bootCPUSpeed() __attribute__((always_inline));
     void static inline bootPowerSaving() __attribute__((always_inline));
 
 


joshgoebel:
This looks good if we can make the changes I suggested to your boards PR I can merge both.

MLXXXp:
Can we mark this as _"Please Hold"_? I want to look into this further when I have more time, but from what I've investigated so far, I'm getting the feeling that attempting any kind of software initiated clock change isn't a good idea (mainly due to USB issues).

This isn't a high priority issue. I think the only advantage running at 8MHz provides is to save power for increased battery life. I'd like some time to form my opinion and recommendations.

joshgoebel:
> I'm getting the feeling that attempting any kind of software initiated clock change isn't a good idea (mainly due to USB issues).

Not that it's not a good idea, just that it requires due diligence.  I probably shouldn't have assumed it would be so simple... :-)

I'll tag it.

MLXXXp:
> Not that it's not a good idea

Sorry, I didn't mean to imply that the idea itself wasn't a good one. I meant that the feasibility of actually being able to implement it without consequences is likely low.

joshgoebel:
> Sorry, I didn't mean to imply that the idea itself wasn't a good one. I meant that the feasibility of actually being able to implement it without consequences is likely low.

Well, sure, consequences of having to fix the USB click.  None of this would be necessary if we owned the boot-loader because then we could have slid this code into a bootloader time setting.  So the clock speed (on the chip) was properly clocked immediately before USB was brought up. :)

MLXXXp:
> None of this would be necessary if we owned the boot-loader

Not true. The USB hardware, including the PLL prescaler, is re-initialised in _main()_ by a call to _USBDevice.attach()_ just before calling the sketch's _setup()_. Any settings done in the bootloader will be overridden by this.

The main issue though, which I haven't had time to fully investigate, is that the USB API allows the ability to _attach_ and _detach_ USB at any time. So, even if we fix the PLL prescaler in our _bootCPUSpeed()_ function, there's no guarantee that it won't be messed up again later by another call to _USBDevice.attach()_.

So far in my investigation, I'm feeling that the only way which would allow safely changing the clock speed through software, in the current Arduino AVR environment, would be to have the Arduino environment decouple the CPU clock speed from the USB clock speed.

I.e.:
There would be another parameter, _build.f_usb_  added to _boards.txt_ for the USB clock speed, which would define F_USB the same way as F_CPU is. The USB core code would then be modified to use F_USB where it is presently using F_CPU. Then, our current 8MHz adjustment code would work as is. For 8MHz, in _boards.txt_ we would put:

```
...build.f_cpu=8000000L
...build.f_usb=16000000L
```

As an aside, I discovered that there already exists a clean, proper mechanism which would allow us to change the CPU clock to 8Mhz before the sketch's _setup()_ is called. You just need to put a _.c_ or _.cpp_ file in your _variants_ directory (where _pins_arduino.h_ goes) containing a function called _initVariant()_.

joshgoebel:
> Not true. The USB hardware, including the PLL prescaler, is re-initialised in main() by a call to USBDevice.attach() just before calling the sketch's setup(). Any settings done in the bootloader will be overridden by this.

True.  The whole idea is to compile for 8Mhz... and then run the chip at 8Mhz... if BOTH those things happen it all works... the problem now is USB code is running at 16Mhz before we can down clock it... IF we owned the bootloader we could down clock is CRAZY early - before ANY other code... so now the chip is acting 8MHz and the code is compiled for 8Mhz.  It would all just work.

joshgoebel:
> As an aside, I discovered that there already exists a clean, proper mechanism which would allow us to change the CPU clock to 8Mhz before the sketch's setup() is called. 

AH, I knew but forgot about this... but it has to be earlier than "before setup"... it has to be before the other device init (like USB)... so I'd have to read the source to see how early they call it... 

So we could move the support from the library to the boards platform?

MLXXXp:
> IF we owned the bootloader we could down clock is CRAZY early - before ANY other code...

Not true. The Leonardo boot loader is already setting up the PLL properly for the physical 16Mhz clock that we're using (as evidenced by the fact that we can always upload code after using the reset switch). It's the call in _main()_ to _USBDevice.attach()_ which improperly changes the PLL prescaler because F_CPU tells it that it's being fed an 8MHz clock.

You have to understand that in the ATmega32U4, the physical clock is split and feeds the CPU clock prescaler separately from the PLL used by the USB. Changing the CPU clock prescaler won't affect the PLL, and vice versa. Because of this, logically it makes sense that there should be a F_USB (or F_PLL) value separate from F_CPU.

> but it has to be earlier than "before setup"... it has to be before the other device init (like USB)

No, due to what I said above, it has to be **after** USB init. Unfortunately, the call to _initVariant()_ is **before** the call to _USBDevice.attach()_.

Here's the Arduino _main()_ function:

``` cpp
int main(void)
{
    init();

    initVariant();

#if defined(USBCON)
    USBDevice.attach();
#endif

    setup();

    for (;;) {
        loop();
        if (serialEventRun) serialEventRun();
    }

    return 0;
}
```

But even if we change the PLL prescaler in _boot()_, after _USBDevice.attach()_ has run from _main()_, my concern is that sometime later, something is going make use of the USB API and call _USBDevice.detach()/USBDevice.attach()_ again, thus changing the PLL prescaler for a 8MHz physical clock (since it's using _F_CPU_ as a guide), and messing up USB again. Even if I don't find any evidence that this is happening, the potential that it can be done means there's no guarantee that it won't be done in the future, or it isn't currently being done by some external library that a sketch may include.

> So we could move the support from the library to the boards platform?

Not as things currently stand. Implementing my proposal for a _F_USB_ define, which could be different from _F_CPU_, would be something that the Arduino IDE itself would have to do. It would be a relatively straight forward change, and be backwards compatible (If a _boards.txt_ file didn't set _f_usb_ it would default to the same as _f_cpu_). It wouldn't be hard to make the changes and offer it as a pull request, but I can't say if or when it would ever be adopted and make it into an IDE release.

joshgoebel:
> Not true. The Leonardo boot loader is already setting up the PLL properly for the physical 16Mhz clock that we're using (as evidenced by the fact that we can always upload code after using the reset switch). 

You're missing my point I think.  The problem is the bootloader is doesn't do anything with clock speed.
If we compiled it ourselves to support BOTH 8 and 16 then the FIRST thing it would do is down clock... and THEN the USB code would expect a certain clock AND receive a certain clock - all works.

Now the USB code is expected 8mhz but getting 16mhz (then getting switched in the middle).  If its was just 8Mhz all along it would all work.

joshgoebel:
Ok, I'll shut up.  I don't have time to read this super carefully... if the USB clock is tied absolutely in some way to the REAL clock then yeah, we might still have a problem - which is what I think you are saying... should still be fixable in software I think... but not as easy as I'd hope.

joshgoebel:
> You have to understand that in the ATmega32U4, the physical clock is split and feeds the CPU clock prescaler separately from the PLL used by the USB. Changing the CPU clock prescaler won't affect the PLL, and vice versa. 

Ok, but that's good.  I'm not sure why initVariant() can't work.  The USB code needs to be compiled with F_CPU 8Mhz then BEFORE it's run the actually clock needs to be 8mhz, which can happen in initVariant... USB expects 8, USB receives 8.

What am I missing still?

joshgoebel:
Unless you're saying the USB clock needs it's own prescaler. :)  And that we also now have to set that to /2. :-)  Then I understand.

MLXXXp:
> Unless you're saying the USB clock needs it's own prescaler.

Correct. There's a separate prescaler in the PLL that provides the USB timing. The PLL is being initialised in _USBDevice.attach()_ based on what F_CPU is set to.

MLXXXp:
> And that we also now have to set that to /2.

Correct; we have to set it for a 16MHz clock **after** _USBDevice.attach()_ has incorrectly changed it for an 8MHz clock. And because the USB API allows calling _USBDevice.attach()_ from within the sketch, it's possible that the sketch would do so, resulting in USB being messed up again.

From USBAPI.h:

``` cpp
class USBDevice_
{
public:
    USBDevice_();
    bool configured();

    void attach();
    void detach();  // Serial port goes down too...
    void poll();
    bool wakeupHost(); // returns false, when wakeup cannot be processed
};
extern USBDevice_ USBDevice;
```

MLXXXp:
I wrote:

> Because of this, logically it makes sense that there should be a F_USB (or F_PLL) value separate from F_CPU.

I'll point out that there's a good case for doing this, for reasons other than fiddling with the CPU prescaler to allow sketches to run at lower speeds. Which means there's a better chance that the Arduino people might actually entertain implementing the idea.

You may want to build an Arduino compatible ATmega32U4 based device that runs at some strange clock speed. Let's say, the ArduPi, which has a 3.1416MHz crystal.

The PLL requires an 8MHz or 16Mhz clock source for USB operation. The ArduPi's clock is 3.1416MHz, so it couldn't be used for USB. However, the ATmega32U4 also contains an _"Internal 8MHz Calibrated Oscillator"_ which could be configured in the bootloader to feed the PLL (allowing USB 1.5MHz Low-Speed operation), while the crystal oscillator could feed the CPU.

This device couldn't work with the current Arduino IDE, but if F_PLL were added (probably a better name since, in addition to USB, it's also used for the High Speed Timer module),  you'd just have to put in _boards.txt_:

```
...build.f_cpu=3141600L
...build.f_pll=8000000L
```

If I get some spare time, I might just try submitting a PR for this, just to see what comes of it. :smile:

MLXXXp:
Superseded by PR #134

PR #113, opened 2016-02-26
Remove logo and clean up ArduBreakout

MLXXXp:
Scroll down "Arduboy" logo was removed because the library now provides a similar logo. I removed the two tone sound at the end of the intro, as well, on the assumption that the built-in library logo will eventually make a sound itself.

setup() and loop() moved to the beginning of the code, as recommended in the [_Arduino style guide_](https://www.arduino.cc/en/Reference/StyleGuide):

> Put your `setup()` and your `loop()` at the beginning of the program. They help beginners to get an overview of the program, since all other functions are called from those two.

Minor changes and cleanup.

### DIFF
--- a/examples/ArduBreakout/ArduBreakout.ino
+++ b/examples/ArduBreakout/ArduBreakout.ino
@@ -9,7 +9,7 @@
  version 2.1 of the License, or (at your option) any later version.
  */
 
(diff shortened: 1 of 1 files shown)


joshgoebel:
Looks good.

PR #115, opened 2016-02-28
fix frame timing

joshgoebel:
previosly timing was calculating the next frame based on now + duration
when the correct timing should have been last_frame + duration.

This would have the effect of slowing the game down the higher your CPU
usage.  A 0% CPU usage game would  play at the request frame rate,
but a 100% CPU usage game would play at only half the requested frame rate.

(RR is render time below)

```
Timing with the bug:
60FPs requested, 35FPs actual, 3 frames, 87ms.
|         29ms         |         29ms         |         29ms         |
| RR 12ms + 17ms sleep | RR 12ms + 17ms sleep | RR 12ms + 17ms sleep |

Timing after the fix:
60FPs, 60FPs actual, 3 frames, 51ms.
|       17ms      |       17ms      |       17ms      |
| RR 12ms + sleep | RR 12ms + sleep | RR 12ms + sleep |
```

After this fix you may need to change your games timing mechanics or
adjust your frame rate (to slow things back down). Your game will run
faster after this fix, even if you were using "100% CPU" before.

### DIFF
--- a/UPGRADING_12.md
+++ b/UPGRADING_12.md
@@ -1,0 +1,35 @@
+Upgrading to 1.2
+================
+
+Version 1.2 of the library includes a few large breaking changes.
+
+### AbPrinter
+
+Arduboy no longer subclasses `Print` so no more printing/text operations.  Use the new AbPrinter class instead.
+
+    // Before
+    Arduboy arduboy;
+    // ...
+    arduboy.print("Hello");
+    
+    // After
+    Arduboy arduboy;
+    // ...
+    AbPrinter text(arduboy);
+    text.print("Hello");
+
+
+### ArduboyTunes
+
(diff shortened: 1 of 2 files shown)


nicolas17:
I think this change made the included ArduBreakout example run way too fast...

joshgoebel:
The example needs to have it's frame rate changed to something that makes the speed seem correct then... the new timing is actually "correct"...  To slow it down the frame rate can just be reduced until it feels right.

MLXXXp:
> The example needs to have it's frame rate changed to something that makes the speed seem correct then...

Yes, that would be the plan. It's just that nobody's gotten around to doing it yet.

I've got it on my list of things to do in the [Arduboy2](https://github.com/MLXXXp/Arduboy2) library I'm working on.

PR #121, opened 2016-03-01
Fix Arduino v1.0.5 compatibility issue with deprecated attribute

dastergon:
Hello,

I am Pavlos from [codebender](https://codebender.cc). We're currently testing your updated library in order to add it in our supported libraries list. The testing showed that there are some issues with the current `Arduboy.h` header.  It looks like the Arduino's 1.0.5 compiler (that codebender is currently compatible with) chokes on `__atributte__((deprecated("message")))`. It supports just `__atributte__((deprecated))` without a message or `__atributte__((warning("message")))` instead. We thought removing the message is better for semantics but if you insist on the message then we could take the `warning` approach.

Please let me know if you any questions.

Thanks,
Pavlos

### DIFF
--- a/src/Arduboy.h
+++ b/src/Arduboy.h
@@ -63,7 +63,7 @@ public:
    */
   // void boot(); // defined in core.cpp
 
-  void start() __attribute__ ((deprecated("use begin() instead")));
+  void start() __attribute__ ((warning("use begin() instead")));
 
   /// Scrolls in the Arduboy logo
   void bootLogo();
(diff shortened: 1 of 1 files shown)


joshgoebel:
Please make these warnings.

joshgoebel:
Actually I read that wrong.  We should use both if this is how it's gotta be:

https://github.com/yyyc514/ArduboyLib/commit/82311b0cc53380527888274b01ce591b40ad734a

Closing this and merging my branch which builds on top of yours.

PR #123, opened 2016-03-02
rewrite deprecated attributes to work with Arduino 1.0.5 tools

joshgoebel:


src/Arduboy.h
```diff
@@ -63,7 +63,7 @@ class Arduboy : public ArduboyCore
    */
   // void boot(); // defined in core.cpp
 
-  void start() __attribute__ ((deprecated("use begin() instead")));
+  void start() __attribute__((deprecated)) __attribute__ ((warning("use begin() instead")));
```
MLXXXp, line 66: I don't think you need to use `__attribute__` twice.
`void start() __attribute__ ((deprecated, warning("use begin() instead")));`

joshgoebel:
Really? If that works submit a PR. :)  Didn' know that.

MLXXXp:
> If that works submit a PR.

Yes, it works. I tried it.

MLXXXp:
Created PR #129.

PR #126, opened 2016-03-02
Backwards compatibility fixes

dastergon:
Hi again,

After some investigation at [codebender](https://codebender.cc) we found out and fixed some compatibility issues between Arduino 1.0.x and the `Arduboy` library.
#### remove duplicate header includes in ArduBreakout

Remove the redundant include
#### include SPI and EEPROM headers explicitly

Although Arduino 1.6.7 identifies the libraries automatically,
older Arduino versions require these libraries to be included explicitly.
#### include Arduino.h file in core.h

Although Arduino 1.6.7 identifies the libraries automatically,
older Arduino versions require these libraries to be included explicitly.
#### tell avr-gcc not to inline fillScreen function

The compile options in Arduino's 105 compiler use the -Os option for optimization
and that has a result the creation of multiple inline functions for fillScreen.
The implications of these inline functions are to make compile in the assembler part.

Let me know if you have any questions.

Pavlos


MLXXXp:
> include SPI and EEPROM headers explicitly

I don't like the idea of doing this. One of the reasons we decided to support only 1.6+ was so that sketches don't need to have this kind of excess useless baggage.

Note that this requirement doesn't apply to just the example sketches in the library. **All** sketches that use the Arduboy library would have to add these includes in order to work on _codebender_

joshgoebel:
Is this the complete list of issues on 1.0.5 or is this only the beginning with more to come?

joshgoebel:
> Note that this requirement doesn't apply to just the example sketches in the library. All sketches that use the Arduboy library would have to add these includes in order to work on codebender

Yes, I agree the SPI and EEPROM stuff is messy, hence my asking if it can be centralized.

shogerr:
> More than 2? How was this discovered? I've never had to tell the compiler not to inline something before 

If you build your own makefile for compiling with avr-gcc you will see this stuff. Then you can goof around with different -On switches.

joshgoebel:
> If you build your own makefile for compiling with avr-gcc you will see this stuff. Then you can goof around with different -On switches.

I'm quite familiar with -O switches (Gentoo, etc).  But I"m not familiar with the compiler randomly inlining things over and over without a good reason.  Typically -oS would do less inlining, not more - hence my asking for any additional context.

joshgoebel:
Doesn't 1.0.5 require all files in a single dir?

dastergon:
@yyyc514 @rogosher  Regarding the multiple inline calls we've written a [PoC](https://gist.github.com/dastergon/f2a73210bfa62c403e86) on how we reproduced and fixed the issue. Let me know if you have any question.

@yyyc514  Regarding the 1.0.5 file structure, indeed it requires `core` and `audio` directories to get merged into a single 'utility' directory. I have pushed an extra commit which uses the 1.0.5 layout and it is also compatible with 1.6.7. 

@yyyc514  Regarding the SPI and EEPROM libraries unfortunately all sketches below 1.6.7 will fail since there's no support for centralised includes.

MLXXXp:
My concern is that if we encourage the Arduboy community to write sketches that can be made available on _codebender_, they won't be able to take advantage of new features and functions provided by the latest IDE.

For instance, they consult the Arduino documentation and find a function (for example, _EEPROM.update()_) and use it in their sketch, only to discover that they get an error when they try compile it on _codebender_.

Or, they find a nice library and use it in their sketch, then they ask _codebender_ to add it, or they [embed it in their sketch](http://feedback.codebender.cc/knowledgebase/articles/92166-uploading-and-including-external-or-custom-librari), only to find that the library uses features only available in IDE 1.6+, so it doesn't compile. The sketch developer may not have the knowledge or time to fix this library themselves, so may end up out of luck.

Most of the Arduboy community will be developing on a 1.6+ version of the Arduino IDE, so they won't discover that their code won't work on _codebender_ until they actually upload it and try to compile. Then they'll say "Why am I getting errors? It works fine in the IDE that I used to develop it.". It looks to me like there may be quite a bit of frustration if we endorse putting Arduboy sketches on _codebender_ , unless _codebender_ can keep pace with the latest Arduino IDE.

joshgoebel:
@MLXXXp Where is your rant about how old 105 is again?

MLXXXp:
@yyyc514, I'm not sure if you're being sarcastic, and I don't believe I've mentioned the age of 1.0.5 (which was released in May 2013), but [yesterday I commented](https://github.com/Arduboy/Arduboy/pull/127#issuecomment-191508507) that Arduino IDE 1.5+ has been worked on since at least [October 2012](https://blog.arduino.cc/2012/10/22/arduino-1-5-support-for-the-due-and-other-processors-easier-library-installation-simplified-board-menu-etc/) and [1.6 has been stable for over a year](https://www.arduino.cc/en/Main/ReleaseNotes).

joshgoebel:
Oh it's 1.!!!5!!! you were talking about.  Yikes.  No, I wasn't being sarcastic at all. :-)

bateske:
@MLXXXp your concern is pretty unfounded, and pretty silly when you think about it.

If we made these changes, users would at least be able to use 1.0.5 or Codebender. Sure some functions might not work but if you don't have a library at all... nothing at all will work.

It would be better to be inclusive rather than exclusive unless it is going to break something in the current codebase.

I'm working on a compatible library because it seems like this is far more a political issue than a technical one, and I actually care about getting into Codebender instead of getting brownie points for bashing an old version of the Arduino library.

The question I have, is where to put the includes for the SPI  and EEPROM within the library to get it so that it will compile. If I put it in the Arduboy.h then core.h complains about not being able to find the SPI.h library. This is the last step I need to get it working. I also don't know how to use github enough to take what I have offline and try to merge it.

shogerr:
Merging PR. Thank you to the codebender team. let's get this merged and keep moving forward. We can trim and prune back after everyone is up and running. This is not a release, so let's work with the solution codebender has kindly provided. Thanks for the discussion, and welcome codebender, to the "democracy" that is the Arduboy PR/issue tracker.

joshgoebel:
Do NOT merge a PR that is still undergoing significant discussion.

MLXXXp:
This means that every library release will have to be checked to make sure it works on _codebender_. The library is now supporting two different platforms, the Arduino 1.6 IDE and _codebender_, and must be written to the lowest common denominator between the two.

The same goes for any user sketches to be hosted on _codebender_. It would be good to put a caution in any relevant user guides or tutorials indicating that compromises may be required if they are planning to place their sketch on _codebender_. Just what are those compromises? The user may not know until they try it because the Arduino documentation generally applies to the latest release and doesn't indicate in which release new functionality became available. At a minimum, sketches will have to include _SPI.h_ and _EEPROM.h_.

Someone should fix the [ArduboyPlaytune](https://github.com/Arduboy/ArduboyPlayTune) library to work on _codebender_ as well.

MLXXXp:
> The question I have, is where to put the includes for the SPI and EEPROM within the library to get it so that it will compile.

You put them in both _Arduboy.h_ and _core.h_ (and _audio.h_, if necessary).

joshgoebel:
@MLXXXp Yes this is a BIGGER deal than just the "right now".  Is code bender always going to give us priority support in the future for EVERY release?  The library is still active and growing.  So far all of the significant contributors seem to be AGAINST making changes to support a VERY VERY old compiler toolchain.  So when we're ready to release a new library is code bender going to always jump right away and test it and fix it for 105?  The fact that they are still on 105 doesn't inspire confidence that they are ready for this type of commitment to stay current.

This matter was already discussed at length: https://github.com/Arduboy/Arduboy/issues/15

The decision was made (with good reasons) to upgrade to 1.6 and the benefits it gives us.  This shouldn't be suddenly reverted because of some pressure from Codebender.

Is Codebender going to apply this same diligence to all the 3rd party libraries being developed for Arudboy - or do they only are about the "official" library?  If they don't plan to tackle the whole thing then the discussion is moot since most of the popular sketch already use other 3rd party libraries.

This is definitely political - it's about what's best for Codebender, rather than Arduboy or it's developer community.  And by "this" I mean the push to change our core library.  There is no reason Codebender can't fork the library to support their platform.  That places the maintenance of the fork on Codebender, where it belongs.  They are responsible for keeping it updated and current to work on their proprietary platform - vs forcing core library developers so support an aging platform we have no way to easily test on and we simply don't care about (no offense).  The core library can continue to support to work on what is already 4 year old technology at this point without being forced to go further back.

If Codebender wants the library to work "as-is" they need to provide a 1.6 toolchain compiler option.

joshgoebel:
Code bender should create and maintain a fork (we can merge it upstream to a branch here regularly) for their uses.  Over time if it makes sense to merge that back in (which would depend on a great many things) then we could pursue that then.

Of course the more time goes on the older that platform is getting, so that is also a consideration.  I think we should be looking forward not backward.  The fork supports those who really really want to use Codebender and accept all the caveats that go along with it - and the core library remains faithful to the rest of us who have been actively contributing.

MLXXXp:
> Code bender should create and maintain a fork (we can merge it upstream to a branch here regularly) for their uses.

Yes, that's a good solution. Arduboy may even wish to host this separate "codebender/1.0.5 compatible" repository themselves. Changes to the main library could be ported to the codebender version to keep it functionally equivalent, as much as possible.

At some point, hopefully soon, codebender would become 1.6 compatible and the fork could be dropped and only the main version would remain.

joshgoebel:
> Yes, that's a good solution. Arduboy may even wish to host this separate "codebender/1.0.5 compatible" repository themselves.

Yep, could also be a separate repository - with git that's not harder.

> At some point, hopefully soon, codebender would become 1.6 compatible and the fork could be dropped and only the main version would remain.

Yes, and at that point all the hard work that's already been doing to make us 1.6 clean doesn't have to be REDONE.  I hate doing things twice.

Issue #118, opened 2016-02-29
Add a #define for the library version


MLXXXp:
A _#define_ should be added, and maintained, in _Arduboy.h_ (or _core.h_ if that makes more sense) that gives the version of the library. I suggest _ARDUBOY_LIB_VER_ as the name. This could be used by sketches for conditionally compiled code, so as to work with multiple non-compatible library versions. 

The format should be the same as the _ARDUINO_ define provided by the IDE to indicate its version:
- The actual version number, in the form of x.y.z, is represented as a single decimal number consisting of:
- The most significant digits are the primary version number of any length, with no leading zeros (leading zeros would be discarded anyway, when used as a numeric value).
- The two next most significant digits are the secondary version number, with a leading zero or just two zeros if necessary.
- The two least significant digits are the tertiary version number, again with a leading zero or just two zeros.

E.g, for version 1.5.42 it would be:
`#define ARDUBOY_LIB_VER 10542`
for 167.59.1
`#define ARDUBOY_LIB_VER 1675901`

For a sketch to work with library version 1.2.0 or higher, and still work with the version it's originally for, it would use:

``` cpp
#include <Arduboy.h>
Arduboy arduboy;
#if defined(ARDUBOY_LIB_VER) && ARDUBOY_LIB_VER >= 10200
AbPrinter text(arduboy);
#endif
// [...]
#if defined(ARDUBOY_LIB_VER) && ARDUBOY_LIB_VER >= 10200
text.print("Hello");
#else
arduboy.print("Hello");
#endif
```

Later, once the newer library version was well established and available, and it was unlikely anyone would still be using the older version, the sketch could be updated to remove the older version compatibility, if desired.

joshgoebel:
I don't think we want to encourage coding to ALL versions of the lib like that, but I'm fine with the version number being defined.  It should go in Ardudoy.h since that is the actual name of the included file and name of the library itself.

Shouldn't we also have `LIB_VER_MAJOR`, `LIB_VER_MINOR`, etc... or is someone suppose to do math if they need to figure that out?

joshgoebel:
In your case it'd make mose sense to do something like this: (change names however)

```
#if defined(ARDUBOY_LIB_VER) && ARDUBOY_LIB_VER >= 10200
AbPrinter text(arduboy);
#define PRINTER text
#endif

PRINTER.print("hi");
```

[...]

PR #127, opened 2016-03-02
Change EEPROM write() to update()

MLXXXp:
Using [_EEPROM.update()_](https://www.arduino.cc/en/Reference/EEPROMUpdate) instead of [_EEPROM.write()_](https://www.arduino.cc/en/Reference/EEPROMWrite) will work exactly the same, except the actual write to EEPROM won't take place if the value already in EEPROM is the same as what's being written. This will prevent impacting the limited number of EEPROM write cycles if there's no need to.

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -20,7 +20,7 @@ void ArduboyAudio::off()
 
 void ArduboyAudio::saveOnOff()
 {
-  EEPROM.write(EEPROM_AUDIO_ON_OFF, audio_enabled);
+  EEPROM.update(EEPROM_AUDIO_ON_OFF, audio_enabled);
 }
 
 void ArduboyAudio::begin()


joshgoebel:
In light of the POTENTIAL effort to make it work [be more compatible] on 1.0.5 is there a way to detect 1.0.5 via compiler define and use write vs update for it still?

MLXXXp:
> on 1.0.5 is there a way to detect 1.0.5 via compiler define and use write vs update for it still?

Yes. You use the same technique as with the _ARDUBOY_LIB_VER_ define we added, described in issue #118.

The _#define_ you test is _ARDUINO_. You just have to determine which version of the IDE added the _EEPROM.update()_ function. From the [_Release Notes_](https://www.arduino.cc/en/Main/ReleaseNotes) this appears to be 1.6.2 but it should probably be confirmed by examining [the repository](https://github.com/arduino/Arduino).

For example, if it actually is 1.6.2 then the code would be"

``` cpp
#if ARDUINO >= 10602
  EEPROM.update(EEPROM_AUDIO_ON_OFF, audio_enabled);
#else
  EEPROM.write(EEPROM_AUDIO_ON_OFF, audio_enabled);
#endif
```

Or, as has been discussed, if you think this function may be used multiple places in the future:
In Arduboy.h or some other header file add:

``` cpp
#if ARDUINO >= 10602
  #define EEPROM_MODIFY update
#else
  #define EEPROM_MODIFY write
#endif
```

And then in the code use:

``` cpp
EEPROM.EEPROM_MODIFY(EEPROM_AUDIO_ON_OFF, audio_enabled);
```

I'm still not happy, though, with having to jump through hoops like this just to make it work with a service that can't get their act together, with Arduino IDE 1.5+ having been worked on since at least [October 2012](https://blog.arduino.cc/2012/10/22/arduino-1-5-support-for-the-due-and-other-processors-easier-library-installation-simplified-board-menu-etc/) and with 1.6 being stable for over a year.

joshgoebel:
> Or, as has been discussed... #define EEPROM_MODIFY update

I was suggesting it in a very specific weird context, I wasn't suggesting we do that everywhere. :)  To me `BLAH.function()` is a little better than `variable.DEFINE_NAME`.

I'm going to hold this one until we figure out what's going on with the 1.0.5 stuff, then we'll see what needs to happen here.

Your dislike of 1.0.5 has been noted. :-)  I'm with you but curious to see what the real effort is.  If it's changing 5 lines I'm more likely to et on board than rearranging the whole repository.

MLXXXp:
We could also just implement the functionality ourselves (and wrap it in a function, if desired), at the risk of a possible increase in code size.

``` cpp
if (EEPROM.read(EEPROM_AUDIO_ON_OFF) != audio_enabled) {
  EEPROM.write(EEPROM_AUDIO_ON_OFF, audio_enabled);
}
```

joshgoebel:
Probably about the same unless we started doing that trick over and over.  Lets wait and see how the 105 bender drama pans out.

joshgoebel:
If we had to support 105 fully I would just consider back porting EEPROM (at least update) into our library. We will see.

MLXXXp:
I guess now that there's a separate _codebender_ branch, this PR can now be applied to the _master_ branch.

PR #129, opened 2016-03-03
Combine function __attribute__ statments

MLXXXp:
Also deleted _getInput_ keyword and added _buttonsState_ keyword.

Note that the `__attribute__` declaration(s) for _getInput()_ have never worked. There was a colon left at the end of _getInput()_, so the `__atribute__` statement wasn't associated with anything and did nothing. I've corrected this.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -14,6 +14,7 @@ Arduboy	KEYWORD1
 
 allPixelsOn	KEYWORD2
 blank	KEYWORD2
+buttonsState	KEYWORD2
 clear	KEYWORD2
 cpuLoad	KEYWORD2
 display	KEYWORD2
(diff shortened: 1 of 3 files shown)


joshgoebel:
Since the whole point of this was 1.0.5 support does this syntax actually confirmed to work on 1.0.5?

Sorry didn't think to ask that earlier.

MLXXXp:
Just tried it with IDE 1.0.5 and, yes, it works.

MLXXXp:
Just tried my IDE 1.0.5 using the original way, with the message as a parameter of _deprecated_:
`void start() __attribute__((deprecated("use begin() instead")));`

It works fine. So we can't trust using an Arduino IDE 1.0.5 as an indication of how _codebender_ is going to behave.

PR #130, opened 2016-03-04
Revert "Backwards compatibility fixes"

joshgoebel:
Reverts Arduboy/Arduboy#126

This has not been properly discussed and all the implications thought thru.  @bateske once said I was in charge of the library... and while I'm not sure if he was joking or not obviously I've been given commit bits and I'm a core contributor.  If we are going to have a few people with such status then we do not just ignore their "Please Hold" tags or that an issue is still undergoing significant discussion.

That is disrespectful and makes me no longer want to be involved.  It was tagged for a reason.

Reverting this patch.


joshgoebel:
I have taken the master branch (before the rivert of the 105 stuff) and pushed it onto the "codebender" branch here.  Future codebender compatibility patches (and updates to keep the library current) should be pull-requests against that branch.

joshgoebel:
This is why github has branches.  Master is `forever` and does NOT need to be polluted with things that are still being actively developed and esp when a lot of contributors have concerns.

PR #132, opened 2016-03-05
Remove duplicate Arduboy.h include from ArduBreakout

MLXXXp:


joshgoebel:
The other examples all prefer "" to <>.

MLXXXp:
`<>` searches the system include list. `""` searches your local directory and then the system include list.
[Include Syntax](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html)

In the IDE, when you use `Sketch -> Include Library -> Arduboy` you will get:

``` cpp
#include <Arduboy.h>
#include <ab_printer.h>
```

MLXXXp:
Using quotes instead of angle brackets probably came about because when the Arduboy DevKit was initially released, examples were putting the Arduboy library in the sketch's folder itself.

MLXXXp:
Added a commit to use angle brackets for Arduboy.h includes in all examples.

PR #134, opened 2016-04-01
Fix setting CPU clock to 8MHz

MLXXXp:
I'm not so sure we want to allow the possibility of changing the clock speed down to 8MHz, but the code currently in the library doesn't quite work properly. This pull request works properly, as far as I've tested it. The main discussion on this is in PR #112 (That PR is superseded by this one and can be ignored/closed). I have a _boards.txt_ file available, that would be required to select 8MHz, if we want to go "all the way" with this feature.

I see some issues with allowing 8MHz operation, which may be reason to just remove the related code completely instead of applying this PR:
- The person compiling the sketch must be the one to decide to run it at 8MHz. The decision to do so would have to be made by trial and error or by the sketch's documentation indicating that it will work properly at the slower speed. There's no way for a sketch itself to force proper 8MHz operation.
- The selected speed has to be checked and possibly changed each time a new sketch is compiled to make sure it's right. If the IDE is left at 8MHz from the previous sketch and then a new sketch is compiled/loaded that requires 16MHz, it will not run properly, possibly causing some "head scratching" or frustration.
- If a person sets for 8MHz but the sketch doesn't use a version of the library that supports it, there's a good chance that the USB port will not function properly after the sketch starts running, requiring the reset button to be pressed to upload a new sketch. This would affect sketches that use their own library, such as those by [Team a.r.g.](http://www.team-arg.org/) (unless they add the equivalent support).
- Even though the code fixes the USB corruption problem when the sketch starts, there's a possibility that a sketch, or a library it uses, could do some USB operations that corrupt the USB port later in the sketch, thus requiring the use of the reset button to recover.
- Probably the only reason to run at 8MHz is to reduce the power used, for longer battery life. No tests have been done to determine how much power would be saved.
- Just the fact of having a selection for 16MHz or 8MHz in the IDE may cause people to ask what it's for and how and when it should be set.

Because of the above, I think most people would probably just always compile for 16MHz, so I don't know how beneficial this feature would be, or whether the benefits outweigh the potential drawbacks.

If we decide it's better to just to remove the 8MHz selection code, I can create a pull request to do so (or maybe just another commit on this one, so the working 8MHz code is stored somewhere for possible future reference). We could also just apply this PR, so it works properly, but not actually make changes to _boards.txt_ to allow it's use.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -93,9 +93,10 @@ ArduboyCore::ArduboyCore() {}
 
 void ArduboyCore::boot()
 {
-  #if F_CPU == 8000000L
-  bootCPUSpeed();
-  #endif
+#ifdef ARDUBOY_SET_CPU_8MHZ
+// ARDUBOY_SET_CPU_8MHZ will be set by the IDE using boards.txt
+  setCPUSpeed8MHz();
+#endif
 
   SPI.begin();
   bootPins();
@@ -109,15 +110,19 @@ void ArduboyCore::boot()
   bootPowerSaving();
 }
 
-#if F_CPU == 8000000L
-// if we're compiling for 8Mhz we need to slow the CPU down because the
-// hardware clock on the Arduboy is 16MHz
-void ArduboyCore::bootCPUSpeed()
+#ifdef ARDUBOY_SET_CPU_8MHZ
+// If we're compiling for 8MHz we need to slow the CPU down because the
+// hardware clock on the Arduboy is 16MHz.
+// We also need to readjust the PLL prescaler because the Arduino USB code
+// likely will have incorrectly set it for an 8MHz hardware clock.
+void ArduboyCore::setCPUSpeed8MHz()
 {
   uint8_t oldSREG = SREG;
   cli();                // suspend interrupts
+  PLLCSR = _BV(PINDIV); // dissable the PLL and set prescale for 16MHz)
   CLKPR = _BV(CLKPCE);  // allow reprogramming clock
   CLKPR = 1;            // set clock divisor to 2 (0b0001)
+  PLLCSR = _BV(PLLE) | _BV(PINDIV); // enable the PLL (with 16MHz prescale)
   SREG = oldSREG;       // restore interrupts
 }
 #endif
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -93,10 +93,10 @@ ArduboyCore::ArduboyCore() {}
 
 void ArduboyCore::boot()
 {
-#ifdef ARDUBOY_SET_CPU_8MHZ
-// ARDUBOY_SET_CPU_8MHZ will be set by the IDE using boards.txt
+  #ifdef ARDUBOY_SET_CPU_8MHZ
+  // ARDUBOY_SET_CPU_8MHZ will be set by the IDE using boards.txt
   setCPUSpeed8MHz();
-#endif
+  #endif
 
   SPI.begin();
   bootPins();
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -283,9 +283,9 @@ protected:
     void static inline safeMode() __attribute__((always_inline));
 
     // internals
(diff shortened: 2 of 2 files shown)


joshgoebel:
If a sketch wants to force 8MHz compile it could just add a compiler error (and warning) if the 8Mhz speed was not defined.  I don't know the syntax to do that but I'd guess it's not that hard.

joshgoebel:
> Even though the code fixes the USB corruption problem when the sketch starts, there's a possibility that a sketch, or a library it uses, could do some USB operations that corrupt the USB port later in the sketch, thus requiring the use of the reset button to recover.

I'm not sure this is saying anything since it's possible at any speed.

> Probably the only reason to run at 8MHz is to reduce the power used, for longer battery life. No tests have been done to determine how much power would be saved.

Yep.  Well and those super paranoid about burning their chip up. :)

> We could also just apply this PR, so it works properly, but not actually make changes to boards.txt to allow it's use.

Lets leave this, esp if you took the time to figure out how to get the USB working at 8Mhz... it might prove valuable one day to someone who really wants it.  At this point if we did rip it out I'd still want to leave it in examples or a doc file or something... to save someone else the time.

joshgoebel:
This is kind of my baby, so hesitant to kill it before anyone has had a change to really use it - and before we have the math on power saving. :-)  If no one cares a year from now and it hasn't been used, then we'll rip it out... :-)

MLXXXp:
> > Even though the code fixes the USB corruption problem when the sketch starts, there's a possibility that a sketch, or a library it uses, could do some USB operations that corrupt the USB port later in the sketch, thus requiring the use of the reset button to recover.
> 
> I'm not sure this is saying anything since it's possible at any speed.

The specific problem I'm inferring is when _F_CPU_ is 8000000L but the true clock speed is 16MHz. If a sketch, or a library it uses, happens to call _USBDevice.attach()_ for some reason (which is perfectly legal), the PLL will be set with the wrong prescaler divider, thus corrupting the USB. This wouldn't happen under normal circumstances, when _F_CPU_ matches the true clock speed.

> > Probably the only reason to run at 8MHz is to reduce the power used, for longer battery life.
> 
> Yep. Well and those super paranoid about burning their chip up.

I don't think I've ever seen any claims that you can physically damage an ATmega32U4, due to overheating, just by running it at 16MHz at a low voltage. It just introduces the possibility of internal timing issues causing code crashes. Besides, If you were super paranoid then you would have to run everything at 8MHz, which would limit you to using only sketches that that run comfortably at this slower clock speed.

> Lets leave this, esp if you took the time to figure out how to get the USB working at 8Mhz...

Fine with me :smile:. I took the time because I didn't like having broken code in the library and knew that there might be objections to removing it completely (plus I find this sort of thing interesting). It doesn't hurt anything if it's there but not used. The only issue might be someone examining the library and wondering what it's purpose is, if we don't add the capability for the user to select the speed.

MLXXXp:
> If no one cares a year from now and it hasn't been used, then we'll rip it out...

Even if we put the selection for it in _boards.txt_, I wonder if we'll be able to track it's usage, since it's something that has to be selected by the person compiling the sketch. The sketch itself can't select it. How will we know who's using it?

I guess it would be better with pre-compiled sketches used with a hex loader, but it still might be difficult to determine which of these sketches were compiled at 8MHz.

joshgoebel:
> I guess it would be better with pre-compiled sketches used with a hex loader, but it still might be difficult to determine which of these sketches were compiled at 8MHz.

No, if we can enforce the use of boards.txt we can encode all sorts of hidden information in a few bytes of the compiled .hex that would be all sorts of useful for a bootloader.

MLXXXp:
If you want to try out a _boards.txt_ file with 16MHz/8MHz selectable support, for both the production Arduboy and DevKit, that works with this PR, add URL
http://www.mlxxxp.ca/arduboy/package_arduboy_index.json
to the IDE Preferences _Additional Boards Manager URLs_ field. Then, search for "arduboy" in
`Tools -> Board -> Boards Manager`

Note that this is just a quick, unofficial package that I may remove in the future.

PR #135, opened 2016-04-13
Added pausing button combo

gnargle:
Adapted from my original code here
https://github.com/gnargle/arduboy-pause-check

I felt that the arduboy needed a specific button combo that could be used as standard across all games, so I've created a function, checkPause that checks all 4 dpad directions and changes a bool if they're all pressed at the same time. 
True = paused, false = not paused.
There's also a second function, setupPause, that relates to how often the bool can be changed.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -18,3 +18,8 @@
   https://github.com/LenShustek/arduino-playtune
   GNU General Public License v3
   (C) Copyright 2011, 2015, Len Shustek
+
+- arduboy-pause-check
+  https://github.com/gnargle/arduboy-pause-check
+  GNU General Public License v3
+  (C) Copyright 2016, Nicholas Allen
--- a/examples/Pausing/Pausing.ino
+++ b/examples/Pausing/Pausing.ino
@@ -1,0 +1,23 @@
+#include "Arduboy.h"
+
+Arduboy arduboy;
+
+bool isPaused = false;
+
+void setup() {
+  // put your setup code here, to run once:
+  arduboy.begin();
+  arduboy.setFrameRate(60);
+  //arduboy.setupPause(1);
+}
+
+void loop() {
+  if (!(arduboy.nextFrame())) return;
+  arduboy.clear();
+  // put your main code here, to run repeatedly:
+  if(isPaused){
+    arduboy.drawCircle(WIDTH/2, HEIGHT/2, 9, WHITE);
+  }
+  isPaused = arduboy.checkPause(isPaused);
+  arduboy.display();
+}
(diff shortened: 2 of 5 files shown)


MLXXXp:
_CONTRIBUTORS.md_, _keywords.txt_ and _Pausing.ino_ have been saved with CR/LF line terminators, not LF terminators.

MLXXXp:
Refactored _checkPause()_:

``` cpp
bool Arduboy::checkPause(bool isPaused)
{
  if (pressed(RIGHT_BUTTON + LEFT_BUTTON + UP_BUTTON + DOWN_BUTTON)
      && (pause_countdown == 0)) {
    isPaused = !isPaused;
    pause_countdown = pause_countdown_set_val;
  }
  if (pause_countdown > 0) {
    pause_countdown--;
  }
  return isPaused;
}
```

Note that _pause_countdown_ is defined as _byte_, which is unsigned. It can never go below 0 so testing for <= 0 is unnecessary.

MLXXXp:
The Arduino environment prefers camelCase. _pause_countdown_ should be _pauseCountdown_ and _pause_countdown_set_val_ should be _pauseCountdownSetVal_

gnargle:
Thanks for the input. If I want to make these changes, do I need to make them in my own repo first then pull request again?

(This is my first ever pull request so I'm not sure on what to do!)

joshgoebel:
This is functionality for individual games to define, this isn't something for the core library.  most games already use a single button for pause/unpause.  This button combo is very strange and hard to press (I know because I used it to trigger soft reboots in by button library).

MLXXXp:
If you change the same branch of your repo, the changes will automatically be added to the pull request.

MLXXXp:
> This is functionality for individual games to define

Although I've been helping out with this, I tend to agree that it should be up to each individual sketch.

We could maybe turn this into an example, though. It demonstrates one method of debouncing buttons that perform toggle actions.

MLXXXp:
> It demonstrates one method of debouncing buttons that perform toggle actions.

However, "_wait for release_" might be a better suited technique for this, in most circumstances.

gnargle:
I appreciate that it's perhaps not the usual thing for the library, but I only really went to the trouble because Kevin suggested I put in a pull request on twitter.
All the same, I appreciate the help on my code.

shogerr:
@gnargle 

> We could maybe turn this into an example, though. It demonstrates one method of debouncing buttons that perform toggle actions.

This is a good suggestion and one I like for this feature request.

A pull request to a public repository usually initiates a discussion about the addition and then a process that involves some editing or refactoring that results in the feature being flatly rejected, accepted, or worked into the library as best our process allows.

I would enjoy your PR as an example personally. That way it is always available to users of the library, but does not dictate the default action of a complex combo of keystrokes.

It's a bummer there is no start button, and it can be added through software if a designer wishes, but the hardware simply doesn't have it, so I think this is why it's outside the core lib's scope.

PR #138, opened 2016-06-01
Add platform.io manifest

fazibear:


fazibear:
Hmm... Someone added it already 2 days ago.

PR #140, opened 2016-06-10
@PlatformIO Library Registry manifest file

valeros:


MLXXXp:
See PR #137 for discussion.

ivankravets:
Thanks, I've just re-registered your library http://platformio.org/lib/show/192/Arduboy

PR #144, opened 2016-07-11
Make color parameter optional. Make data types consistent.

MLXXXp:


joshgoebel:
Just FYI: This could break sketches that link against us and call these functions with the older types.

MLXXXp:
The types don't change, just the names. They are _typedef_ed to the same data types internally somewhere along the line.

joshgoebel:
`unsigned char` and `uint8_t` are not the same type.  Try making a function and passing a variable of type unsigned char when the function expects uint8_t.

joshgoebel:
This came up all the time when working with images since some arrays were unsigned char and some uint8_t.

joshgoebel:
If you use raw numbers in your code you don't tend to notice since C will do a lot of the munging for you but when you use variables and arrays it gets messier.

MLXXXp:
> unsigned char and uint8_t are not the same type.

_unsigned char_ has been changed to _uint8_t_ in three places: function _drawChar()_, the screen buffer _sBuffer_ and function _getBuffer()_.

_drawChar()_ is called by _write()_ whose prototype has always been:

`virtual size_t write(uint8_t);`

So it should have been a _uint8_t_ in _drawChar()_ in the first place. In V1.1 it's _unsigned char_, which hasn't caused any problems.

I've switched _sBuffer_ to _uint8_t_ in my _Arduboy2_ library. It hasn't caused any problems with any of the sketches I've ported to it.

As an oversight, I didn't change _getBuffer()_ in _Arduboy2_, so there's a mismatch. This hasn't caused any problems in tests I've done. Changing _getBuffer()_ to _uint8_t_, to match _sBuffer_, hasn't caused problems either.

You can have a function prototype declare a _uint8_t_ and then use _unsigned char_ in the function definition and it works. e.g.:

``` cpp

// prototype in file test.h:
class Test
{
 public:
  void myFunction(uint8_t x);
};

// definition in file test.cpp
void Test::myFunction(unsigned char x)
{
  Serial.print(x);
}
```

If _uint8_t_ and _unsigned char_ aren't equivalent, why doesn't the above cause a compile error when _myFunction()_ is called?

My efforts to find a difference between _unsigned char_ and _uint8_t_ have failed. Perhaps you can provide an example where changing from V1.1 to the current _develop_ branch would break.

MLXXXp:
I found a post that gives a technique to find the basic data type of any data type, by the error that the compiler produces:
http://stackoverflow.com/questions/34920697/show-most-basic-type-declaration-of-nested-typedefs-in-c/34920865#34920865

I made a sketch that incorporates this:

``` cpp
// Find the basic data type for any data type
template<typename> struct TC;
template<typename> using A = uint8_t; // <- type that you want to find the basic type for
using B = A<float>;
void setup() {TC<B> a;}
void loop() {}
```

When you compile this you get:
`aggregate 'TC<unsigned char> a' has incomplete type and cannot be defined`

This confirms that  _uint8_t_ and _unsigned char_ **are** the same type in the Arduino AVR environment.

joshgoebel:
```
uint8_t blah = 0;

void test(unsigned char a) {
}

void main() {
  test(blah)
}
```

I'm saying (from my recollection) that this will not compile without an error, am I mistaken?

MLXXXp:
Turning it into a sketch, and adding some code to ensure _test()_ isn't optimised out of the link, it compiles fine, with no errors or warnings.

``` cpp
uint8_t blah = 0;

void setup() {
  test(blah);
}

void loop() {}

void test(unsigned char a) {
  Serial.print(a);
}
```

joshgoebel:
Weird.  Ah, maybe it's only arrays or pointers to arrays that are an issue then...   I know I'm not imagining it.  Or maybe the compiler toolchain has improved since 6 months ago?

MLXXXp:
Maybe you're thinking of `const` vs non-constants, or PROGMEM mismatches?

MLXXXp:
> Or maybe the compiler toolchain has improved since 6 months ago?

I just tried the above sketch with IDE V1.6.7, released Dec 17, 2015, and also the old IDE V1.0.5, without any errors or warnings.

I also tried reversing it (with both older releases); passing _unsigned char blah = 0;_ to _test(uint8_t a)_ and still no problem.

I also tried passing an array element (with both older releases and with the types reversed); _uint8_t blah[] = {0, 1};_ and _unsigned int blah[] = {0, 1};_. I still had no problems at all.

PR #145, opened 2016-07-18
Keep V1.1 nextFrame() and add newFrame(). Put beginNoLogo() back in

MLXXXp:
These changes are to help maintain backwards compatibility with the V1.1 API
- Leave the timing bug in _nextFrame()_ but deprecate the function. This way, existing sketches using _nextFrame()_ will continue to run at the same speed. The version of _nextFrame()_ with the bug fix is added in and named _newFrame()_. All new development should use _newFrame()_. Existing sketches can be modified to use it if desired.
- Put an equivalent version of the V1.1 _beginNoLogo()_ function back in. This will allow existing sketches using it to compile. New development should use _boot()_ plus, optionally, additional functions from _begin()_.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -51,7 +51,7 @@ height	KEYWORD2
 idle	KEYWORD2
 initRandomSeed	KEYWORD2
 invert	KEYWORD2
-nextFrame	KEYWORD2
+newFrame	KEYWORD2
 notPressed	KEYWORD2
 off	KEYWORD2
 on	KEYWORD2
--- a/src/Arduboy.cpp
+++ b/src/Arduboy.cpp
@@ -8,7 +8,7 @@
 #include "ab_logo.c"
 #include "glcdfont.c"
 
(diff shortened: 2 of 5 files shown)


MLXXXp:
Added a commit to put the deprecated function _getInput()_ back in. We're trying to make this as backwards compatible to V1.1 as possible.

I think this function was unintentionally removed instead of leaving it in but letting it remain deprecated.

MLXXXp:
According to a comment made by @dastergon in PR #123 the compiler used by Arduino V1.0.5 doesn't like a _deprecated_ attribute with a message, which causes problems with _codebender_. However, I've tried it with V1.0.5 and found that _deprecated_ attributes with a message work properly.

There's no need to use a _warning_ attribute to provide a message, so a commit has been added to this PR to remove them.

Besides, if this is a problem with _codebender_ then it can be addressed in the _ArduboyClasssic_ library that _codebender_ uses.

PR #148, opened 2016-07-26
Put getInput() back in. Remove warning attributes

MLXXXp:
This is a merge-able version of the last two commits of PR #145: 

---

Added a commit to put the deprecated function _getInput()_ back in. We're trying to make this as backwards compatible to V1.1 as possible.

I think this function was unintentionally removed instead of leaving it in but letting it remain deprecated.

---

According to a comment made by @dastergon in PR #123 the compiler used by Arduino V1.0.5 doesn't like a _deprecated_ attribute with a message, which causes problems with _codebender_. However, I've tried it with V1.0.5 and found that _deprecated_ attributes with a message work properly.

There's no need to use a _warning_ attribute to provide a message, so a commit has been added to this PR to remove them.

Besides, if this is a problem with _codebender_ then it can be addressed in the _ArduboyClasssic_ library that _codebender_ uses.

### DIFF
--- a/library.properties
+++ b/library.properties
@@ -7,3 +7,4 @@ paragraph=This library is for content creation on the Arduboy, a portable gaming
 category=Other
 url=https://github.com/arduboy/arduboy
 architectures=avr
+includes=Arduboy.h
--- a/src/Arduboy.h
+++ b/src/Arduboy.h
@@ -99,7 +99,7 @@ public:
   void begin();
 
   /// Depreciated function. Use begin instead.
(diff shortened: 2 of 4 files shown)


MLXXXp:
In the Arduino IDE as of V1.6.10 there is a new "includes" property for file _library.properties_. This specifies which files to include in a sketch when IDE menu item `Sketch -> Include Library` is used.

A commit has been added to this PR to set the _includes_ property to _Arduboy.h_. This will prevent _ArduboyCore.h_, _ArduboyAudio.h_ and any other future unwanted _.h_ files from being included.

PR #149, opened 2016-08-01
Fix paintScreen() for IDE 1.6.10

MLXXXp:
Arduino IDE 1.6.10 uses a new version of the compiler toolchain. The compiler now generates more efficient code so that the instruction padding in function _paintScreen()_, used to create a delay so a byte can be shifted out over SPI, is no longer sufficient.

The problem could be solved by adding more pad instructions, but this doesn't eliminate the possibility that the same thing won't occur in the future.

This PR fixes the problem by waiting for the _SPIF_ flag to indicate that the byte has been sent and the SPI controller is ready for a new one, instead of blindly using a delay to allow time for the byte to be sent.

This technique takes a bit longer to update the display. Previously the total time was 1.16 ms and now it's 1.42 ms (which is still shorter than using _SPI.transfer()_ for each byte, which takes 2.58 ms). 

Although it appears to be a significant increase in time, overall it shouldn't have too much of an impact. A sketch that used 100% of the CPU time to render frames at 60 fps will now still run at 59 fps.

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,14 @@
+{
+    "name": "Arduboy",
+    "keywords": "arduboy, game",
+    "description": "This library is for content creation on the Arduboy, a portable gaming platform",
+    "repository":
+    {
+        "type": "git",
+        "url": "https://github.com/Arduboy/Arduboy.git"
+    },
+    "version": "1.1.1",
+    "exclude": "extras",
+    "frameworks": "arduino",
+    "platforms": "atmelavr"
+}
--- a/library.properties
+++ b/library.properties
@@ -1,5 +1,5 @@
 name=Arduboy
-version=1.1.0
+version=1.1.1
 author=Chris J. Martinez, Kevin Bates, Josh Goebel, Scott Allen, Ross O. Shoger
 maintainer=Ross O. Shoger ros@arduboy.com
 sentence=The Arduboy core library.
@@ -7,3 +7,4 @@ paragraph=This library is for content creation on the Arduboy, a portable gaming
 category=Other
 url=https://github.com/arduboy/arduboy
 architectures=avr
+includes=Arduboy.h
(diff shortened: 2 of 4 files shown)


shogerr:
We should leave this issue until this is properly tagged and pushed at the IDE release list. What are the ramifications of this change for users of older IDES?

Can we push these changes at the previous releases? Or is there a reason to maintain the assembly for this function in older versions, for use by older IDEs.

MLXXXp:
With my changes merged, we can tag this release immediately. It should be tagged **1.1.1** (with no "**v**" prefix).
I can do this, if you wish.

However, I think _README.md_ should first be tweaked just a bit. The parts about board selection, using examples, running on a DevKit, and maybe more, should be removed and just refer to the Quick Start Guide in the forum. There's no need to maintain this information in two places.

> Can we push these changes at the previous releases?

You should never push changes to a previous release without changing the version number. Once a release is tagged it must remain locked in stone.

> Or is there a reason to maintain the assembly for this function in older versions, for use by older IDEs.

The changes will work with all  older IDEs. As I said, it will take slightly longer than before to update the screen, but it won't have any impact unless the sketch is currently using 100% of the CPU time available to render each frame (and even then I doubt the 1.5% slowdown would be noticeable).

MLXXXp:
> However, I think README.md should first be tweaked just a bit.

OK, maybe more that just a bit.  :grin: I've created PR #150 with changes to README.md. If this is acceptable, merge it, and then I think we should release as 1.1.1 so people can update their IDE to 1.6.10

Once released, Arduino will then automatically pick it up shortly and make it available in the IDE.

I can both merge README.md and create the release, if you wish.

ace-dent:
@MLXXXp as this replaces the ASM by @yyyc514 I wondered if you benchmarked how the compiled size has increased?

MLXXXp:
@ace-dent, I think the size increases by about 12 bytes.

If you care about how I benchmarked the timings:
I wrote a sketch that sets Arduino Pin 2 high and then calls _display()_ and then sets Pin 2 low. I attached an oscilloscope to Pin 2 and measured the time between it going high and then low.

Issue #89, opened 2016-02-16
What would we name a combined displayclearbuffer?


joshgoebel:
We need to add this for performant games... we lose a lot of time with SPI and one way to get it back is to do the clear for the next frame while we paint the current frame to the hardware.  That gives us almost 1% CPU performance back, or more time to sleep and save power.

I'm just not sure what to call it.

`displayClearBuffer()`
`displayAndClear()`
`paintAndClear()`

Thoughts?

joshgoebel:
We have about 10-15,000 process cycles we're wasting doing nothing while the slow SPI does it's thing... if we can think of any other housekeeping that can be done incrementally during screen painting we could maybe shove it in here.  It just has to be doable in tiny chunks of time.

joshgoebel:
That's for EACH and every screen refresh... At 60 FPS we lose like almost 1Mhz of our 16Mhz just WAITING on the screen, not even doing anything. :-)  I think I just figured out where to hide the greyscale code if we ever got 2 bit color working. :-)

shogerr:
`displayFast()`? I know we talked about _paint_ as a way to express drawing to the screen, to avoid some issues. So if we move to that, then `paintFast()` seems appropriate as well.

MLXXXp:
displayThenClear()

Assuming it will produce the same result as:

``` cpp
arduboy.display();
arduboy.clear();
```

ghost:
I would like to test it on Shadow Runner using arduboy.cpuLoad() to see the difference.

joshgoebel:
> displayThenClear()

I like it.  Not 100% clear regarding what is happening, but effectively true as far as the results.

Actually in order to understand it it really needs to be `displayAndClear()`... the other way isn't clear and makes you wonder why we're creating one big function that just runs one after the other.

joshgoebel:
> I would like to test it on Shadow Runner using arduboy.cpuLoad() to see the difference.

You might not be able to.  cpuLoad math is all done in ms... the precision isn't necessarily high enough to detect a 1% improvement.  You'd have to measure microseconds before and after.  That's what I've used for accurate measurement in the past.  cpuLoad would be better done in microseconds too but that would take up extra RAM and flash for something not helpful to most users.

joshgoebel:
cpuLoad was always intended as more of a high level estimation tool, not to be used for precision measurements of differences between algorithms.  It will tell you if you're in the right ballpark, but not if one thing is 2% faster than another.

[...]

PR #152, opened 2016-08-11
Fix paintScreen for IDE 1.6.10 and add combo display/clear functions

MLXXXp:
Main commits:
- Applied the _paintScreen()_ fix from V1.1.1 which fixes the screen corruption problem with Arduino IDE 1.6.10
- Added core function _paintScreenAndClearImage()_, used by new function _displayAndClear()_. Many sketches call _display()_ and then use _clear()_ before rendering the next frame. Providing a function to do both saves these sketches a function call. Mainly though, because the buffer bytes are cleared while waiting for them to shifted out to the display, it doesn't take any more time than just calling _display()_. You save the extra time that a separate _clear()_ call would take.
  This was proposed in issue #89, which could be closed if this PR is applied.

Minor commits:
- For function _fillScreen()_, changed the default value for the _color_ argument from WHITE to BLACK. I think clearing the screen buffer to black (all zeros) will likely be the most common use of _fillScreen()_. This won't affect any existing code since nothing is currently calling _fillScreen()_ with a default.
- Removed the unnecessary use of the _this->_ pointer in function _drawBitmap()_. It's not used anywhere else. The library coding style should be consistent.

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -23,6 +23,7 @@ clear	KEYWORD2
 cpuLoad	KEYWORD2
 digitalWriteRGB	KEYWORD2
 display	KEYWORD2
+displayAndClear	KEYWORD2
 drawBitmap	KEYWORD2
 drawChar	KEYWORD2
 drawCircle	KEYWORD2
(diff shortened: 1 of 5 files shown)


shogerr:
There may be a cleaner way to execute the clears and draws and for the default value of function calls, specifically _fillScreen()_ in this case, is possible. I assume a call with on, not off. Pushing the commit. Thanks MLXXXp.

PR #154, opened 2016-08-26
Add dummy tunes functions. Change paintScreen() and display() related functions.

MLXXXp:
The ArduboyTunes class was removed to allow external sound libraries to be used.
Its functions have been reimplemented as deprecated dummy functions so that
existing sketches will compile without errors (but will produce no sound).
Working versions of the functions are available in the ArduboyPlaytune library.


src/ArduboyCore.cpp
```diff
@@ -242,54 +242,33 @@ void ArduboyCore::paintScreen(const uint8_t *image)
 
 // paint from a memory buffer, this should be FAST as it's likely what
 // will be used by any buffer based subclass.
-// if this function is changed, make sure corresponding changes
-// are made to paintScreenAndClearImage()
-void ArduboyCore::paintScreen(uint8_t image[])
+void ArduboyCore::paintScreen(uint8_t image[], bool clear)
```
shogerr, line 245: We probably want to set a default, (..., bool clear = 0)

MLXXXp: A default of _false_ **is** set in the function prototype in _ArduboyCore.h_. You set default values in the prototype, not the definition.

PR #155, opened 2016-09-11
Fix: drawFastHLine checking uint8_t for negative value

Joshimuz:
drawFastHLine was using uint8_t w for negative value checks, however
instead the value was underflowing and causing graphical issues and
buffer overflow crashes when called for a position/width off screen.


Joshimuz:
http://community.arduboy.com/t/the-bounce-a-ball-physics-platformer/2144/4

MLXXXp:
I've evaluated this fix and an equivalent one I wrote using a different algorithm. Mine produces less code and runs faster, so it was added as PR #156 instead of this using this one.

@Joshimuz, thanks for submitting this pull request, even though it wasn't used.

